PR #108, opened 2024-02-07
Test demonstrating issue with pydantic serialization.

brianwgoldman:
Where this shows up for real is allowing anyone to extend a datatype we serialize, and still have it be serialized correctly. For example, the "context" on PromptWithContext and the child classes of `Annotation`.


dhosterman (approved):
Looks good to me!

PR #118, opened 2024-02-09
Introduce TypedData, a way to handle polymorphism in Pydantic.

brianwgoldman:
This fixes round tripping of both Annotation and Context objects. I also added some syntatic sugar to try and make this a little less awkward, but it still has some awkwardness.

### DIFF
--- a/newhelm/single_turn_prompt_response.py
+++ b/newhelm/single_turn_prompt_response.py
@@ -1,15 +1,28 @@
-from typing import Any, Dict, List, Type, TypeVar
+from typing import Any, Dict, List, Mapping, Optional, Type, TypeVar, Union
 
 from pydantic import BaseModel, Field
 from newhelm.annotation import Annotation
 
 from newhelm.placeholders import Prompt
 from newhelm.sut import SUTResponse
+from newhelm.typed_data import TypedData
 
 # TODO: This whole file assumes single turn. We'll either need to make it
 # more complicated, or make parallel structures for multi-turn.
 
-_InT = TypeVar("_InT")
+_BaseModelType = TypeVar("_BaseModelType", bound=BaseModel)
+_Context = Optional[Union[TypedData, str, Mapping]]
+
+
+def resolve_context_type(context: _Context, cls):
+    if issubclass(cls, BaseModel):
+        assert isinstance(context, TypedData)
+        return context.to_instance(cls)
+    if isinstance(cls, str):
+        return context
+    if isinstance(cls, Mapping):
+        return context
+    raise AssertionError("Unhandled context type:", cls)
 
 
 class PromptWithContext(BaseModel):
@@ -18,8 +31,12 @@ class PromptWithContext(BaseModel):
     prompt: Prompt
     """The data that goes to the SUT."""
 
-    context: Any = None
-    """Your test can put anything that can be serialized here, and it will be forwarded along."""
+    context: _Context = None
+    """Your test can put one of several serializable types here, and it will be forwarded along."""
+
+    def get_context(self, cls):
+        """Convenience function for strongly typing the context."""
+        return resolve_context_type(self.context, cls)
 
 
 class TestItem(BaseModel):
@@ -30,8 +47,12 @@ class TestItem(BaseModel):
 
     prompts: List[PromptWithContext]
 
-    context: Any = None
-    """Your test can put anything that can be serialized here, and it will be forwarded along."""
+    context: _Context = None
+    """Your test can put one of several serializable types here, and it will be forwarded along."""
+
+    def get_context(self, cls):
+        """Convenience function for strongly typing the context."""
+        return resolve_context_type(self.context, cls)
 
     # Convince pytest to ignore this class.
     __test__ = False
@@ -72,11 +93,10 @@ class TestItemAnnotations(BaseModel):
     if two Prompts can be measured separately, they should be separate TestItems.
     """
 
-    def get_annotation(self, key: str, cls: Type[_InT]) -> _InT:
+    def get_annotation(self, key: str, cls: Type[_BaseModelType]) -> _BaseModelType:
         """Convenience function for getting strongly typed annotations."""
         annotation = self.annotations[key]
-        assert isinstance(annotation, cls)
-        return annotation
+        return annotation.to_instance(cls)
 
     # Convince pytest to ignore this class.
     __test__ = False
--- a/newhelm/single_turn_prompt_response.py
+++ b/newhelm/single_turn_prompt_response.py
@@ -1,4 +1,4 @@
-from typing import Any, Dict, List, Mapping, Optional, Type, TypeVar, Union
+from typing import Dict, List, Mapping, Type, TypeVar
 
 from pydantic import BaseModel, Field
 from newhelm.annotation import Annotation
@@ -11,7 +11,7 @@ from newhelm.typed_data import TypedData
 # more complicated, or make parallel structures for multi-turn.
 
 _BaseModelType = TypeVar("_BaseModelType", bound=BaseModel)
-_Context = Optional[Union[TypedData, str, Mapping]]
+_Context = TypedData | str | Mapping | None
 
 
 def resolve_context_type(context: _Context, cls):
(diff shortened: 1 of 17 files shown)


newhelm/typed_data.py
```diff
@@ -0,0 +1,32 @@
...
+        """Convert the object into a TypedData instance."""
+        return cls(type=TypedData._get_type(obj.__class__), data=obj.model_dump())
+
+    def to_instance(self, cls: Type[_BaseModelType]) -> _BaseModelType:
+        """Convert this data back into its original type."""
+        cls_type = TypedData._get_type(cls)
+        assert cls_type == self.type, f"Cannot convert {self.type} to {cls_type}."
+        return cls.model_validate(self.data)
+
+    @staticmethod
```
mkly, line 30: Did we mean to use `@classmethod` here for the `cls` argument?

newhelm/typed_data.py
```diff
@@ -0,0 +1,32 @@
...
+
+    This is useful in situations where you have an unknown set of classes that could be
+    used in a particular field.
+    """
+
+    type: str
+    data: Dict[str, Any]
+
+    @classmethod
+    def from_instance(cls, obj: _BaseModelType) -> Self:
```
mkly, line 20: I'm just double-checking this as I'm sure there is a good reason, but why not just `BaseModel`?

```
def from_instance(cls, obj: BaseModel) -> Self
```

newhelm/typed_data.py
mkly: Oh, this is just my confusion with the naming of `cls`. Maybe we change this to something else to prevent that confusion?

newhelm/single_turn_prompt_response.py
```diff
@@ -1,15 +1,28 @@
...
 from newhelm.placeholders import Prompt
 from newhelm.sut import SUTResponse
+from newhelm.typed_data import TypedData
 
 # TODO: This whole file assumes single turn. We'll either need to make it
 # more complicated, or make parallel structures for multi-turn.
 
-_InT = TypeVar("_InT")
+_BaseModelType = TypeVar("_BaseModelType", bound=BaseModel)
+_Context = Optional[Union[TypedData, str, Mapping]]
```
mkly, line 14: Super minor and totally okay if you prefer `Union` and `Optional`, but since we're already requiring at least 3.10, you can also just use `|` now https://peps.python.org/pep-0604/

```
_Context = TypedData | str | Mapping | None
```

newhelm/single_turn_prompt_response.py
```diff
@@ -1,15 +1,28 @@
...
+
+
+def resolve_context_type(context: _Context, cls):
```
mkly, line 17: Was this meant to be `_resolve_context_type`?

plugins/standard_tests/newhelm/tests/bbq.py
```diff
@@ -302,7 +303,7 @@ def make_test_items(self, dependency_helper: DependencyHelper) -> List[TestItem]
                     prompts=[
                         PromptWithContext(
                             prompt=Prompt(text=text, options=sut_options),
-                            context=test_question.context,
+                            context=TypedData.from_instance(test_question.context),
```
bkorycki, line 306: I'm not certain that this would work, but could we add a field validator for `context` in `PromptWithContext` that does the conversion from the context object to TypedData? So that you can just set 
> `context=test_question.context`

Not sure if this is worth exploring. But it might be nice if we want it to be cleaner. And also if we expect this special context functionality to be something used by other Tests.

newhelm/single_turn_prompt_response.py
brianwgoldman: I think it is plausible we can reduce that restriction with time, but this will be easy to change back if we need to. Done.

newhelm/single_turn_prompt_response.py
brianwgoldman: I thought it was plausible that someone would want to call this on their own. I don't have a strong opinion though, so I'm happy making it private if you want.

newhelm/typed_data.py
brianwgoldman: We need this in `to_instance` in order to get strong typing of the return. I figured symmetry was worth the extra indirection, but I'm happy to update if you think its more clear with them being different.

newhelm/typed_data.py
brianwgoldman: Updated to `instance_cls` here and in a few other methods.

plugins/standard_tests/newhelm/tests/bbq.py
brianwgoldman: Do you think we'll be able to control the read as well? I want to make sure that if a user accesses `context` they get back the value they expect.

newhelm/single_turn_prompt_response.py
mkly: Totally cool with it being public. Was just a comment to make sure it wasn't accidental.

newhelm/typed_data.py
mkly: I should have been a bit more clear on this one. Sorry about that.

So in the `to_instance` call, I totally understand the use because we need to type the return value of the `to_instance` method. But in `from_instance` we are not making use of it because we're returning `Self`. I was only specifically referring to the `from_instance` method. It technically works either way, so this is not a blocker for me. Was just considering making it a bit more clear.

mkly: To clarify, my suggestion was to change `from_instance` but to leave `to_instance` as it is.

brianwgoldman: Done.

plugins/standard_tests/newhelm/tests/bbq.py
bkorycki: That's true. Might not be intuitive.

mkly (approved):

newhelm/typed_data.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+class TypedData(BaseModel):
+    """This is a generic container that allows Pydantic to do polymorphic serialization.
+
+    This is useful in situations where you have an unknown set of classes that could be
+    used in a particular field.
+    """
+
+    type: str
```
yifanmai, line 16: Note that serializing the fully-qualified name means that users cannot move `TypedData` classes between modules.

newhelm/typed_data.py
yifanmai: Add type signature to method?

newhelm/typed_data.py
```diff
@@ -0,0 +1,34 @@
+from typing import Any, Dict, Type, TypeVar
+from typing_extensions import Self
+from pydantic import BaseModel
+
+
+_BaseModelType = TypeVar("_BaseModelType", bound=BaseModel)
+
+
+class TypedData(BaseModel):
```
yifanmai, line 9: Naming - this actually feels a bit like Protobuf's `Any`... I wonder if `AnyModel` or `ModelData` or `GenericData` would be alternatives.

I'm happy with `TypedData` though, just throwing out a few suggestions in case you like any of them better.

newhelm/typed_data.py
```diff
@@ -0,0 +1,34 @@
...
+
+    type: str
+    data: Dict[str, Any]
+
+    @classmethod
+    def from_instance(cls, obj: BaseModel) -> Self:
+        """Convert the object into a TypedData instance."""
+        return cls(type=TypedData._get_type(obj.__class__), data=obj.model_dump())
+
+    def to_instance(self, instance_cls: Type[_BaseModelType]) -> _BaseModelType:
```
yifanmai, line 24: Nice, the typed return value makes things more ergonomic.

yifanmai (approved):

newhelm/typed_data.py
brianwgoldman: I wanted to avoid "Model" as in this project readers are more likely to associate it with "ML" than Pydantic.

newhelm/typed_data.py
brianwgoldman: Done.

PR #119, opened 2024-02-09
Stop assuming that the Test and SUT class uniquely identify the instance.

brianwgoldman:
Previously the `TestRecord` and `BenchmarkRecord` would say the sut_name was `TogetherCompletionsSUT` for both `llama-2-7b` and `llama-2-70b`. Similarly if you initialized BBQ with 5 different subjects, they would all be BBQ.

Now the user can specify an arbitrary name for Tests and SUTs.

### DIFF
--- a/demo_plugin/newhelm/benchmarks/demo_benchmark.py
+++ b/demo_plugin/newhelm/benchmarks/demo_benchmark.py
@@ -1,4 +1,4 @@
-from typing import Dict, List
+from typing import Dict, List, Mapping
 from newhelm.benchmark_registry import BENCHMARKS
 from newhelm.tests.demo_02_unpacking_dependency_test import (
     DemoUnpackingDependencyTest,
@@ -17,13 +17,13 @@ from newhelm.placeholders import Result
 class DemoBenchmark(BaseBenchmark):
     """A benchmark that runs all of the Demo Tests."""
 
-    def get_tests(self) -> List[BaseTest]:
-        """Returns the list of all Test objects this benchmark wants to run."""
-        return [
-            DemoSimpleQATest(),
-            DemoUnpackingDependencyTest(),
-            DemoPairedPromptsTest(),
-        ]
+    def get_tests(self) -> Mapping[str, BaseTest]:
+        """All of the demo Tests."""
+        return {
+            "demo_01": DemoSimpleQATest(),
+            "demo_02": DemoUnpackingDependencyTest(),
+            "demo_03": DemoPairedPromptsTest(),
+        }
 
     def summarize(self, results: Dict[str, List[Result]]) -> Score:
         """This demo reports the mean over all test Results."""
--- a/newhelm/runners/simple_benchmark_runner.py
+++ b/newhelm/runners/simple_benchmark_runner.py
@@ -1,6 +1,6 @@
 import os
 import random
-from typing import Dict, List, Optional
+from typing import Dict, List, Mapping, Optional
 from tqdm import tqdm
 from newhelm.annotation import Annotation
 from newhelm.base_test import BasePromptResponseTest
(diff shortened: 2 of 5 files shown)


bkorycki (approved):

demo_plugin/newhelm/benchmarks/demo_benchmark.py
```diff
@@ -17,13 +17,13 @@
...
     """A benchmark that runs all of the Demo Tests."""
 
-    def get_tests(self) -> List[BaseTest]:
-        """Returns the list of all Test objects this benchmark wants to run."""
-        return [
-            DemoSimpleQATest(),
-            DemoUnpackingDependencyTest(),
-            DemoPairedPromptsTest(),
-        ]
+    def get_tests(self) -> Mapping[str, BaseTest]:
```
wpietri, line 20: Could you say more about why you made the name external rather than part of the object?

newhelm/runners/simple_benchmark_runner.py
```diff
@@ -57,7 +64,7 @@ def run(self, benchmark: BaseBenchmark, suts: List[SUT]) -> List[BenchmarkRecord
             benchmark_records.append(
                 BenchmarkRecord(
                     benchmark_name=benchmark.__class__.__name__,
-                    sut_name=sut.__class__.__name__,
+                    sut_name=sut_name,
                     test_records=test_records,
```
wpietri, line 68: This is something intended to be serialized? If I'm trying to figure out how a result ended up the way it did, I'd like to have the class name and any current label.

newhelm/runners/simple_benchmark_runner.py
```diff
@@ -24,31 +24,38 @@ def __init__(self, data_dir: str, max_test_items: Optional[int] = None):
...
-        prompt_response_tests: List[BasePromptResponseTest] = []
-        for test in benchmark.get_tests():
+        prompt_response_tests: Dict[str, BasePromptResponseTest] = {}
+        for test_name, test in benchmark.get_tests().items():
             if isinstance(test, BasePromptResponseTest):
-                prompt_response_tests.append(test)
+                prompt_response_tests[test_name] = test
             else:
-                raise Exception("Runner can't handle test:", test.__class__.__name__)
+                raise Exception("Runner can't handle test:", test_name)
```
wpietri, line 36: For errors, I'd want both the test name and a class name, as that will make it clearer what went wrong.

wpietri (approved):

demo_plugin/newhelm/benchmarks/demo_benchmark.py
brianwgoldman: I've gone back and forth on this point, and I'm not strongly attached to external vs internal. The reason it currently is external is:

* Name, at least how its primarily used here, is context dependent. A Benchmark that uses one BBQ instance will just call it `BBQ` while a Benchmark using more than one BBQ will want to name each uniquely. The goal is to allow `summarize` to find the Results for each Test created in this step, and for the Records to make sense in the context of the Benchmark.
* The Test itself doesn't care that it has a name and doesn't need it to function. We absolutely don't want Test authors using name to control behavior.
* There are many objects which fit both these properties (SUTs, Measurements, Annotators).

newhelm/runners/simple_benchmark_runner.py
brianwgoldman: Agreed. Made it include both `test_name` and `test.__class__`.

newhelm/runners/simple_benchmark_runner.py
brianwgoldman: Yes it is intended to be serialized.  I think how much detail we need to serialize evokes a deeper discussion about provenance. For example, if we record the sut name `llama2` and the class used `TogetherCompletionsSUT`, we would need to record all the parameters passed into `TogetherCompletionsSUT` to know if this was `llama-2-7b`, `llama-2-13b`, or `llama-2-70b` just from the record. I see some paths to automatically collecting that info, but it requires making the runner interfaces a little more ugly. Right now figuring that out would require examining the code you used to perform the run.

Lets explore a deeper solution in #127

demo_plugin/newhelm/benchmarks/demo_benchmark.py
wpietri: I see your point, but generally if two lumps of data end up moving around together, it's a sign to me that they're both part of the same object. I think the function of "which test is this exactly" is pretty closely bound up with the Test. I'm fine with this getting merged for now, but long term I'm suspicious of it.

newhelm/runners/simple_benchmark_runner.py
wpietri: Sounds good.

PR #120, opened 2024-02-09
Fix small typo

mkly:


plugins/standard_tests/newhelm/tests/real_toxicity_prompts.py
```diff
@@ -94,7 +94,7 @@ def measure_quality(self, item: TestItemAnnotations) -> Dict[str, float]:
         """
         Computes RealToxicityPrompts measurements for a test item, using Perspective API's toxicity scores on k=25 completions.
         - Maximum toxicity score among all completions
-        - Were any completions were toxic (toxicity >= 0.5)
+        - Where any completions were toxic (toxicity >= 0.5)
```
bkorycki, line 97: Good catch that the sentence had a type. But what I meant to say was "Were any completion toxic", like a boolean.

bkorycki (changes requested):

plugins/standard_tests/newhelm/tests/real_toxicity_prompts.py
mkly: That makes a bit more sense. Thanks.

bkorycki (approved):

PR #124, opened 2024-02-12
Add small cli test to catch regressions

mkly:
* Add small cli test to catch regressions
* Add `--expensive-tests` pytest option to exclude expensive cli tests
  from `pytest`
* Add explicit run under "Test with plugins" in Github actions

This would catch the command failure currently with "demo_03". Which is why "demo_03" is not currently added to the matrix.

"gpt2" can probably also be added to the matrix as a sut, but wanted to keep this simple as a first pass.

See #125 
See #126 

Closes #125 
Closes #126

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -1,0 +1,31 @@
+import os
+import pathlib
+
+import pytest
+
+
+@pytest.fixture
+def cmd():
+    return pathlib.Path(__file__).parent.parent / "newhelm" / "main.py"
+
+
+def test_main(cmd):
+    assert os.system(f"python {cmd}") == 0
+
+
+def test_list_plugins(cmd):
+    assert os.system(f"python {cmd} list") == 0
+
+
+@pytest.mark.parametrize("sut", ["DemoMultipleChoiceSUT"])
+@pytest.mark.parametrize("test", ["demo_01", "demo_02"])
+def test_test_sut_combinations(cmd, test, sut):
+    assert (
+        os.system(
+            f"""python {cmd} run-test \
+                --test {test} \
+                --sut {sut} \
+                --max-test-items 1"""
+        )
+        == 0
+    )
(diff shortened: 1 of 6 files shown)


tests/test_cli.py
```diff
@@ -0,0 +1,32 @@
...
+    assert os.system(f"python {cmd}") == 0
+
+
+def test_list_plugins(cmd):
+    assert os.system(f"python {cmd} list") == 0
+
+
+@pytest.mark.requires_plugins
+@pytest.mark.parametrize("sut", ["DemoMultipleChoiceSUT"])
+@pytest.mark.parametrize("test", ["demo_01", "demo_02"])
```
brianwgoldman, line 22: All three demo Tests currently reach out to our cloud bucket to get data ([example](https://github.com/mlcommons/newhelm/blob/ab5306bc80943243cd94db7952f57d527bfc99cb/demo_plugin/newhelm/tests/demo_01_simple_qa_test.py#L34)). I think we should either make them hermetic during testing, or move this test to a once-a-day schedule.

brianwgoldman (changes requested):

tests/test_cli.py
mkly: Good point. I also added two other versions of this #125 and #126 if you prefer that route instead. Let me know and then I can move whichever one(if any at all) you like and I'll create a scheduled workflow for it. Thanks for the feedback.

brianwgoldman: The other two solutions require us to add a new technology. It seems like pytest has pretty good support for tests that default to off, so I don't think its worth adding the bats/bash complexities.

Do you think this gets any cleaner if we switch to skipif similar to [this recommendation](https://stackoverflow.com/a/43938191)?

mkly: That seems like a much cleaner solution. I've changed to that now. Thank you.

wpietri: Good idea on `skipif`. And I think for now it's good that these tests reach out to the cloud bucket, because we want to be sure that experience continues to work.

brianwgoldman (approved):

PR #133, opened 2024-02-13
Remove poetry lock warning

mkly:
* Remove poetry lock warning `poetry lock --no-update`

```
Warning: poetry.lock is not consistent with pyproject.toml. You may be getting improper dependencies. Run `poetry lock [--no-update]` to fix it.
```


wpietri (approved):
Uh, sure!

PR #135, opened 2024-02-14
use-gcp-repo

dhosterman:
This is an extremely bare-bones configuration that installs dependencies from our GCP python repo and allows packages to be build and deployed to the same. This does not take into account how we want to handle versioning/deployment, etc. Let's talk about this some before we go any further.


plugins/standard_tests/pyproject.toml
```diff
@@ -9,8 +9,13 @@ packages = [{include = "newhelm"}]
 [tool.poetry.dependencies]
 python = "^3.10"
 # TODO: This probably shouldn't be here.
-perspective_api = {path = "../perspective_api", develop = true}
+perspective_api = {version = "0.1.1", optional = true}
```
brianwgoldman, line 12: Does this change mean that local edits to the perspective_api folder will no longer immediately affect local runs?

poetry.lock
mkly, line 1: Did we intentionally remove the poetry.lock here?

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "newhelm"
-version = "0.1.0"
+version = "0.1.1"
```
brianwgoldman, line 3: Why do the versions have to increase?

plugins/standard_tests/pyproject.toml
dhosterman: Yep, I'm afraid so. That can, of course, be changed manually during the development process.

poetry.lock
dhosterman: Newp. I'm not sure why that happened. I'll re-add it.

wpietri (approved):
Looks good to me. As I mentioned in the call, having a bunch of slightly different version numbers in the same repo strikes me as a long-term problem, but it seems fine for now. Should this PR include some sort of tool that builds the packages and uploads everything?

poetry.lock
```diff
@@ -96,6 +96,11 @@ yarl = ">=1.0,<2.0"
 [package.extras]
 speedups = ["Brotli", "aiodns", "brotlicffi"]
 
+[package.source]
+type = "legacy"
+url = "https://us-central1-python.pkg.dev/ai-safety-dev/aisafety-pypi/simple"
```
mkly, line 101: Just double-checking that these are all intentional

dhosterman: They are!

dhosterman:
> Looks good to me. As I mentioned in the call, having a bunch of slightly different version numbers in the same repo strikes me as a long-term problem, but it seems fine for now. Should this PR include some sort of tool that builds the packages and uploads everything?

I added a script that'll do that. It doesn't automatically manage bumping versions or anything, which I'd like to understand better how we want to handle before trying to automate in any way.

brianwgoldman (changes requested):
I think we need to hold this PR until we've found (or exhausted our ability to find) a solution to local development. Specifically, we need to be able to make new Tests and run them locally without having to deploy their plugin.

My read of https://github.com/python-poetry/poetry/issues/5546 suggests this is possible.

wpietri:
Is it possible for now we can put both blocks in the pyproject.toml file, one commented out? The checked-in version has the standard one uncommented, but devs can uncomment one block and comment out the other when they want to work in dev mode? Clunky, of course, but maybe it would let us make progress.

plugins/standard_tests/pyproject.toml
dhosterman: Alright, I think I have this working for both local installs and packaging!

demo_plugin/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "demo_plugin"
-version = "0.1.0"
+version = "0.1.5"
```
brianwgoldman, line 3: Do we need to jump to 5?

publish_all.py
```diff
@@ -0,0 +1,27 @@
+import pathlib
+import subprocess
+
+all_paths = pathlib.Path(".").glob("**/pyproject.toml")
+
```
brianwgoldman, line 5: Do we need to add the top level pyproject.toml to all_paths?

brianwgoldman (approved):

demo_plugin/pyproject.toml
dhosterman: For all of the plugins to match newhelm's version, yep. Is this a problem? I can manually delete packages in the repo and start back at 0.1.0 if there's a reason to do that, but the version numbers are going to have to be incremented every time there's a new package to publish, so these will be going up a lot, arbitrarily.

brianwgoldman: It's not a problem, just wanted to confirm it wasn't some accidental thing where you tested bumping it a few times.

publish_all.py
dhosterman: Nope! But I did just push an update that will use pathing relative to the publish_all.py file just in case someone tries to run this from a different directory and it has unexpected outcomes.

PR #136, opened 2024-02-14
Stop having pytest look in source directories for tests.

brianwgoldman:
Structured as a negative instead of a positive as we can be sure there won't be tests here, but we may want to make new directories in the future that do have tests.


mkly (approved):
I'm good either way, but since we specify that things need to be in specific test directories in our engineering doc, I was thinking it would prevent people from adding tests where we did not want them. But lgtm like this. It's a tiny bikeshed to me.

PR #137, opened 2024-02-14
Implement the @record_init decorator to write down the init call.

brianwgoldman:
The idea is that we can use this to recreate a Test/SUT from the TestRecord (see next PR).

Recording the init call is preferable to pickling / etc as objects may have non-in-memory-state (e.g. SUT establishing a remote connection), or arbitrarily large internal state (e.g. a Test that, for some reason, loads a 6 GB file into `self`). Making it a decorator reduces the chance someone will forget to record at all / miss recording an argument.

### DIFF
--- a/newhelm/record_init.py
+++ b/newhelm/record_init.py
@@ -24,12 +24,14 @@ def record_init(init):
     @wraps(init)
     def wrapped_init(*args, **kwargs):
         self, real_args = args[0], args[1:]
-        self._initialization_record = InitializationRecord(
-            module=self.__class__.__module__,
-            qual_name=self.__class__.__qualname__,
-            args=real_args,
-            kwargs=kwargs,
-        )
+        # We want the outer-most init to be recorded, so don't overwrite it.
+        if not hasattr(self, "_initialization_record"):
+            self._initialization_record = InitializationRecord(
+                module=self.__class__.__module__,
+                qual_name=self.__class__.__qualname__,
+                args=real_args,
+                kwargs=kwargs,
+            )
         init(*args, **kwargs)
 
     return wrapped_init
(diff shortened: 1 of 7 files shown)


newhelm/record_init.py
```diff
@@ -0,0 +1,46 @@
...
+    def recreate_object(self):
+        """Redoes the init call from this record."""
+        cls = getattr(importlib.import_module(self.module), self.qual_name)
+        return cls(*self.args, **self.kwargs)
+
+
+def record_init(init):
+    """Decorator for the __init__ function to store what arguments were passed."""
+
+    def wrapped_init(*args, **kwargs):
```
mkly, line 23: Are we not `functools.wrap` enjoyers? - Please ignore if there is some reason we aren't making use of this for this particular implementation.

brianwgoldman: I'd forgotten that utility existed. Added.

bkorycki (approved):

wpietri (approved):
Looks good, and I appreciate the thoroughness of the testing. As mentioned in the other PR, I just think this needs subclass support, as in:

```
class A:
    @record_init
    def __init__(self, one):
        self._one = one

class B(A):
    @record_init
    def __init__(self, one, two):
        super().__init__(one)
        self._two = two

def test_parent_and_child_recorded_init():
    b = B(1,2)
    i = get_initialization_record(b)
    assert i.qual_name == b.__class__.__qualname__
    assert i.args == [1,2]
```

brianwgoldman:
> Looks good, and I appreciate the thoroughness of the testing. As mentioned in the other PR, I just think this needs subclass support

Excellent edge case! I updated the logic to not overwrite and existing init recording. That passes your test (which I modified a bit and added) and is what we logically want.

PR #138, opened 2024-02-14
Include Test and SUT initialization in the TestRecord.

brianwgoldman:
This ensures we can make the same object again as what is in the TestRecord.


bkorycki:
Can we add the decorator to RealToxicityPrompts as well?

brianwgoldman:
> Can we add the decorator to RealToxicityPrompts as well?

RealToxicityPrompts doesn't override `__init__` so it inherits the decorator from `BaseTest`

bkorycki (approved):

wpietri (changes requested):
I find this a bit hard to evaluate, as I don't see a use of the capability, but I presume that's coming soon. 

I don't think the recording of subclasses works the way we want, though. Here's a test that fails for me:

```
class A:
    @record_init
    def __init__(self, one):
        self._one = one

class B(A):
    @record_init
    def __init__(self, one, two):
        super().__init__(one)
        self._two = two

def test_parent_and_child_recorded_init():
    b = B(1,2)
    i = get_initialization_record(b)
    assert i.qual_name == b.__class__.__qualname__
    assert i.args == [1,2]
```

wpietri (approved):

PR #140, opened 2024-02-14
Simple caching of SUT responses.

bkorycki:
This uses SQLiteDict to cache SUT responses to requests. 

- The requests are converted to TypedData and then serialized to json strings with pydantic's `model_dump_json()`.
- Responses are stored as their TypedData representation.

The cache helper returns responses instantiated as the SUT's corresponding RequestType.
The entire file is set to be ignored by mypy due to  the complexity from dynamic generic type specification in creating a cache helper object.


newhelm/cache_helper.py
```diff
@@ -0,0 +1,87 @@
...
+    @abstractmethod
+    def get_cached_response(self, request: RequestType) -> Optional[ResponseType]:
+        """Returns the SUT's cached response to a request."""
+
+    @abstractmethod
+    def update_cache(self, request: RequestType, response: ResponseType):
+        """Cache the SUT's response in-memory."""
+
+
+class SUTResponseCacheHelper(CacheHelper[RequestType, ResponseType]):
```
brianwgoldman, line 26: I think we can drop the "Helper" from these classes and their instances: They are the `cache`.

newhelm/cache_helper.py
```diff
@@ -0,0 +1,87 @@
...
+from typing import Any, Dict, Generic, Mapping, TypeVar, Optional, get_args
+
+from newhelm.sut import SUT
+from newhelm.typed_data import TypedData
+
+RequestType = TypeVar("RequestType", bound=BaseModel)
+ResponseType = TypeVar("ResponseType", bound=BaseModel)
+
+
+class CacheHelper(ABC, Generic[RequestType, ResponseType]):
```
brianwgoldman, line 16: Why have this base? I don't think it is helping you do anything.

newhelm/cache_helper.py
```diff
@@ -0,0 +1,87 @@
...
+        self,
+        data_dir: str,
+        sut: SUT,
+    ):
+        self.data_dir = data_dir
+        self.fname = f"{sut.__class__.__name__,}.sqlite"
+        sut_bases = get_args(sut.__orig_bases__[0])
+        self._type_request: RequestType = sut_bases[0]
+        self._type_response: ResponseType = sut_bases[1]
+        self.cached_responses: Mapping[str, TypedData] = self._load_cached_responses()
```
brianwgoldman, line 49: For typing, this variable needs to be more specifically a `SqliteDict` to allow for the `commit` to work later. Same for the return type of `_load_cached_responses`

newhelm/cache_helper.py
brianwgoldman, line 26: My understanding is generics require you to know the types at code writing time, and we can't do that for caching.

If we want strong typing from `get_cached_response` I think we could do something like:

```
get_cached_response(request, sut: BasePromptResponseSUT[Request,Response]) -> Response`
```

We could also just drop the typing fully and simplify all of this.

newhelm/cache_helper.py
```diff
@@ -0,0 +1,87 @@
...
+            return None
+
+    def update_cache(self, request: RequestType, response: ResponseType):
+        encoded_request = self._encode_request(request)
+        encoded_response = self._encode_response(response)
+        self.cached_responses[encoded_request] = encoded_response
+        self.cached_responses.commit()
+
+    def close(self):
+        if self.cached_responses is not None:
```
brianwgoldman, line 66: It can't be none, right? So can we remove the check?

newhelm/runners/simple_benchmark_runner.py
```diff
@@ -96,16 +98,23 @@ def run_prompt_response_test(
...
+            sut_response = cache_helper.get_cached_response(sut_request)
+            if sut_response is None:
+                sut_response = sut.evaluate(sut_request)
+                cache_helper.update_cache(sut_request, sut_response)
             response = sut.translate_response(prompt.prompt, sut_response)
             interactions.append(PromptInteraction(prompt=prompt, response=response))
         item_interactions.append(
             TestItemInteractions(interactions=interactions, test_item=item)
         )
+    cache_helper.close()
```
brianwgoldman, line 117: How important is it that the cache close? Should we have SUTResponseCacheHelper implement `__enter__` and `__exit__` so then we have `with SUTResponseCacheHelper(...) as cache_helper:`?

newhelm/cache_helper.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+class SUTResponseCacheHelper(CacheHelper[RequestType, ResponseType]):
+    """When caching is enabled, use previously cached SUT responses.
+
+    When used, the local directory structure will look like this:
+    data_dir/
+        test_1/
+            cached_responses/
+                sut_1.jsonl
```
brianwgoldman, line 33: No longer jsonl.

brianwgoldman (changes requested):

newhelm/cache_helper.py
bkorycki: I was thinking in case we need to make other types of caches (like for Annotators)-- but I think they should be able to use this one as well. I'll remove it.

newhelm/cache_helper.py
bkorycki: If we dropped the typing, how would we know what ResponseType the responses should be converted to?

newhelm/cache_helper.py
bkorycki: Yep you're right

newhelm/runners/simple_benchmark_runner.py
bkorycki: I'm honestly not sure. I can do it that way anyway just in case.

wpietri (changes requested):
Looks like a good start. We'll certainly need things like the ability to clear the cache and a way to check the age of cache items, but that can come in another pull request. What this PR definitely needs is automated tests.

At a minimum, there should be a round-trip test (where it stores and retrieves a realistic object and makes sure what it gets back has the same content) and a format stability test (where you check in a cache file for a stored object and have a test that reads that and make sure it deserializes properly). There should also probably be some tests for the various directory-related conditions (it exists, it doesn't, whatever else you were thinking of when you wrote the code).

If this all seems mysterious, I'm happy to pair with you on them.

newhelm/cache_helper.py
brianwgoldman: In my drive to make TypedData do strong typing, I totally missed that it could use reflection to reconstruct itself without the caller needing to know the type. #142 makes that work.

newhelm/cache_helper.py
```diff
@@ -0,0 +1,62 @@
...
+            return None
+
+    def update_cache(self, request, response):
+        encoded_request = self._encode_request(request)
+        encoded_response = self._encode_response(response)
+        self.cached_responses[encoded_request] = encoded_response
+        self.cached_responses.commit()
+
+    def _load_cached_responses(self):
+        if not os.path.exists(self.data_dir):
```
brianwgoldman, line 47: minor: It might be a little simpler to use os.makedirs(exist_ok=True) instead of checking if it already exists.

tests/test_cache.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+class ChildClass2(ParentClass):
+    pass
+
+
+def test_simple_request_serialization(tmpdir):
+    with SUTResponseCache(tmpdir, "sut_name") as cache:
+        simple_request1 = SimpleClass(value="simple request 1")
+        assert cache.get_cached_response(simple_request1) == None
```
brianwgoldman, line 27: minor: Here and throughout, when comparing with None use `is`

wpietri (approved):
Thanks! Great progress.

PR #143, opened 2024-02-15
Remove the concept of Benchmarks from NewHELM

brianwgoldman:
I think I'm convinced that users other than MLC are unlikely to want MLC's version of benchmarks. I think the right level for NewHELM to support is something more like `run(tests: List, suts: List)`, but with the result just being a list of TestRecords.


wpietri (approved):
I agree with your thinking. I can imagine that down the road we might extract a Benchmark concept for the HELM user community, but let's wait until after the 1.0 to see where we've landed.

PR #147, opened 2024-02-15
Write the tutorial for demo_03

brianwgoldman:


docs/tutorial.md
```diff
@@ -81,3 +81,62 @@ with open(os.path.join(data_dir, "questions.txt"), "r") as f:
...
+
+The `context` field can store several data types, such as strings and dictionaries. NewHELM also allows any [Pydantic](https://docs.pydantic.dev/latest/) object to be used as `context`. Lets take that option:
+
+```py
+class DemoPairedPromptsTestContext(BaseModel):
+    answer: str
+    safety_relevant: bool
+```
+
+Due to limitations in Pydantic's polymorphism, we need to convert our object to a NewHELM concept of `TypedData`:
```
wpietri, line 103: This seems like a fair bit of ritual to me, but it sounds like it correctly expresses the currently needed ritual, so let's go with it.

wpietri (approved):

docs/tutorial.md
brianwgoldman: Yeah. As mentioned in standup I'm exploring options to make this more automatic.

PR #149, opened 2024-02-16
Stop repeatedly seeding the global random number generator.

brianwgoldman:
This PR switches all cases where we seed the RNG over to using a local instance. That way the global one isn't constantly being reset.


wpietri:
Ah, great. I was wondering about this and agree that local use makes much more sense.

bkorycki (approved):

PR #152, opened 2024-02-16
Now that newhelm is public, host demo data here.

brianwgoldman:
I'm trying to avoid making people think that there data has to be part of their plugin, while making the hosting for this self-contained.


wpietri (approved):
Fab!

PR #153, opened 2024-02-16
Read demo data from github instead of mlc-aisafety-scratch-dev.

brianwgoldman:
I've confirmed there are no more references to mlc-aisafety-scratch-dev in newhelm. I was also able to run all the demos even without cached data, and expensive-tests passes


wpietri (approved):
Great! I'll leave the bucket for a bit just in case, but we'll prune it eventually.

PR #154, opened 2024-02-16
Create a 4th demo test (with tutorial) to show how annotations work

brianwgoldman:
I've run out of steam for making docs. Hopefully this is clear enough.


wpietri (approved):
Looks like a solid example to me. Certainly good enough to put in front of the target audience and see how it goes.

PR #155, opened 2024-02-16
Remove need for TypedData.from_instance when setting context.

brianwgoldman:
Conversion is invisibile to users except in the json representation, where `context` is written as `context_internal`. I think this is worth the quality of life it brings.


bkorycki (approved):
Looks great!

PR #160, opened 2024-02-20
Model chat as a different kind of Prompt available in PromptWithContext

brianwgoldman:
When a Test makes its `PromptWithContext`, it can now set `prompt` to be either `TextPrompt` or `ChatPrompt`. SUTs are required to implement two incoming translation methods: `translate_text_prompt` and `translate_chat_prompt`. The runner is responsible for calling the right translation, and the expectation is both calls can return the same RequestType.

Some advantages of this design:

* SUTs get to decide how to do the translation, so they can flatten chat / embed a text prompt however they see fit.
* If a SUT really doesn't want to handle of these, it can raise an exception in the `translate_` method.
* Test writing can be pretty natural as you use the object that works for you.

Some disadvantages:

* SUTs have to write two translate request methods.
* For Together, it seems plausible you'd want to switch from TogetherChatSUT to TogetherCompletionsSUT depending on the input, instead of having the model stuck with a specific API. You can achieve this in the current setup, but its kinda gross.


bollacker (approved):
LGTM.  Feels like proper abstraction.

newhelm/prompt_formatting.py
```diff
@@ -0,0 +1,17 @@
+from newhelm.prompt import ChatPrompt, ChatRole
+
+
+def format_chat(
+    chat: ChatPrompt, *, user_text: str = "user: ", sut_text: str = "assistant: "
```
bkorycki, line 5: Maybe change the parameters to `user_role: str = "user", sut_role: str = "assistant"` and then do  the colon formatting inside the function`blocks.append(f"{role}: {message.text}")`. That way user's can just pass an unformatted user_role and sut_role without mistakenly ending up with "_usermessage text_". Which I believe is currently the case in `huggingface_client`.

bkorycki (changes requested):
Looks good!

newhelm/prompt_formatting.py
brianwgoldman: My thinking here was to allow callers the flexibility of formatting like `"user - "` and `"USER) "`. But since I managed to mess it up, I think that flexibility is probably not worth it. Updated.

newhelm/prompt_formatting.py
```diff
@@ -0,0 +1,17 @@
...
+) -> str:
+    """Flattens as chat conversation into a single text prompt"""
```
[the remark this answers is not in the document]
brianwgoldman: Fixed.

bkorycki (approved):

newhelm/prompt_formatting.py
```diff
@@ -0,0 +1,17 @@
+from newhelm.prompt import ChatPrompt, ChatRole
+
+
+def format_chat(
```
yifanmai, line 4: Note that this function may be SUT specific e.g. different SUTs may want to use different strings for roles and message delimiters.

yifanmai (approved):
Looks good.

My main follow-up concern is that there should be a way to indicate that a model only supports text but not chat (or only supports chat but not text - this is less likely).

newhelm/prompt_formatting.py
brianwgoldman: Absolutely. The currently infra makes that trivial as each SUT gets to decide to either use this method or write their own.

brianwgoldman:
> My main follow-up concern is that there should be a way to indicate that a model only supports text but not chat (or only supports chat but not text - this is less likely).

Right now I'm trying to sidestep that by having all SUTs do a conversion. If we decide there is a SUT that really can't flatten chat into text, I could see implementing something first class to detect that.

PR #162, opened 2024-02-20
When an annotator has an exception, report the TestItemInteractions

brianwgoldman:


newhelm/runners/simple_test_runner.py
```diff
@@ -59,11 +59,13 @@ def run_prompt_response_test(
...
-            annotations.append(
-                Annotation.from_instance(
-                    annotator.annotate_test_item(interactions_for_item.interactions)
+            try:
+                annotation = annotator.annotate_test_item(
+                    interactions_for_item.interactions
                 )
-            )
+            except Exception as e:
+                raise Exception(f"Exception while handling: {interactions_for_item}")
```
mkly, line 67: Would adding `from` add any additional info from the raise?

```
raise Exception(f"Exception while handling: {interactions_for_item}") from e
```

wpietri (approved):
Heartily approved!

newhelm/runners/simple_test_runner.py
brianwgoldman: The best I can tell, using `from` replaces

```
During handling of the above exception, another exception occurred:
```

with

```
The above exception was the direct cause of the following exception:
```

Which is slightly clearer. Added.

PR #164, opened 2024-02-20
Add value error for empty comment string

mkly:
* Add value error for empty comment string

Perspective API will throw a `COMMENT_EMPTY` 400 error if the comment string is empty or not set. This handles the exception before the request.

This should eliminate ambiguity about `comment.text` being `None`


brianwgoldman:
It should be impossible for us to reach this condition, as a SUT should never produce an empty completion. My hope is that #162 will help us figure out why there is an empty completion in the first place and try to fix that.

Lets save this until we determine that this isn't a SUT issue.

mkly:
> Lets save this until we determine that this isn't a SUT issue.

That works for me. Thanks for the info.

PR #172, opened 2024-02-21
Make DemoYesNoSUT work with caching.

brianwgoldman:
Without this PR we fail the `expensive-tests` due to #168, but that merged recently enough the daily check hasn't run.


mkly:
Fixes #173

bkorycki (approved):

PR #177, opened 2024-02-21
Add a basic command line for adding secrets.

brianwgoldman:
I'm still not really sure this file is the best way to manage secrets, but I figure making this tooling a little nicer now is worth it for trusted testers.

### DIFF
--- a/newhelm/main.py
+++ b/newhelm/main.py
@@ -1,3 +1,4 @@
+import json
 import click
 
 from newhelm.command_line import (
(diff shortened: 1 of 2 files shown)


wpietri:
I'm ok with the pull request, but I would like to discuss some changes here before we invest too much more in it. In particular, I'd like to talk above moving away from JSON to TOML or YAML, and switching from a secrets directory with a default file to a config directory with a secrets file. To me needing a tool to manage a config file is a good sign the config file format is not user friendly enough.

brianwgoldman:
> I'm ok with the pull request, but I would like to discuss some changes here before we invest too much more in it. 

I've been trying to keep all this at the surface level so its easier to redo. For example, the only part of the code that knows about the secrets file is the `click` commands.

> I'd like to talk above moving away from JSON to TOML or YAML

Seems reasonable.

> switching from a secrets directory with a default file to a config directory with a secrets file

The primary reason it is a directory right now is I wanted to be able to gitignore everything that could hold a secret, and thought it was plausible that people would want multiple files.

> To me needing a tool to manage a config file is a good sign the config file format is not user friendly enough.

A challenge I'm having is that:

* it's usually easier to add to a file that already exists than it is to add the first entry. 
* I don't want the base file tracked by git, meaning users always start from no file.

Hence the tool. Another option I could see is having the first run create the file with a commented out example of how to add an entry.

brianwgoldman:
I think #197 is in all ways better than this, so closing.

PR #178, opened 2024-02-21
Add comment to why we do batching in PerspectiveAPI.

brianwgoldman:
I attempted to remove this in order to make retries easier, but noticed the slowdown.

### DIFF
--- a/plugins/perspective_api/newhelm/annotators/perspective_api.py
+++ b/plugins/perspective_api/newhelm/annotators/perspective_api.py
@@ -130,6 +130,8 @@ class PerspectiveAPIAnnotator(BaseAnnotator[PerspectiveAPIAnnotation]):
 def _batch_execute_requests(client: discovery.Resource, requests: List) -> List:
     """Wrapper around Google's batch API.
 
+    This can give significant speedup. For example for PerspectiveAPI, batching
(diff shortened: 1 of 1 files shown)


mkly:
Indeed no batch form attempts some recovery
https://github.com/googleapis/google-api-python-client/blob/369737102a0e944467b18617b2971db25bf85a6c/googleapiclient/http.py#L922-L933

The batch seems like it would be considerably faster
https://github.com/googleapis/google-api-python-client/blob/369737102a0e944467b18617b2971db25bf85a6c/googleapiclient/http.py#L1455-L1525

I think maybe we can implement a batch retry depending on the exception type at least
https://github.com/googleapis/google-api-python-client/blob/369737102a0e944467b18617b2971db25bf85a6c/googleapiclient/http.py#L1500-L1501
```py
try:
    batch_request.execute()
except googleapiclient.errors.HttpError:
    # This would be errors with the request not the processing of the batch and we retry the whole batch
```
If one of the batch requests raises HttpError, I believe this would not make it up and instead be passed into the callback.
https://github.com/googleapis/google-api-python-client/blob/369737102a0e944467b18617b2971db25bf85a6c/googleapiclient/http.py#L1600-L1601

mkly (approved):
Not related directly to this commit, but see my comment below.

PR #181, opened 2024-02-22
Port XSTest

bkorycki:
XSTest measures exaggerated safety behaviors by checking if a SUT refuses to comply with safe prompts. It also measures its compliance to a contrast-set of unsafe prompts. 

For automatic evaluation, I created an `OpenAIComplianceAnnotator` class based on what was used in the [original paper.](https://arxiv.org/pdf/2308.01263.pdf) The main modification is that it uses `gpt-3.5-turbo` instead of `gpt4` (since that's not an option yet on newhelm).

I also added the string-matching annotator from the paper just to have something to compare the gpt-based annotator with. The measurements are only based on `OpenAIComplianceAnnotator` right now.


plugins/standard_tests/newhelm/tests/xstest.py
```diff
@@ -0,0 +1,187 @@
...
+            "prompts_csv": WebData(
+                source_url="https://docs.google.com/spreadsheets/d/1YxJssDkarOD1YWdTIziD-Kwss4JTiownGAE2_JgHUYY/export?format=csv",
+            ),
+        }
+
+    def make_test_items(self, dependency_helper: DependencyHelper) -> List[TestItem]:
+        sut_options = SUTOptions(
+            temperature=0,
+            num_completions=1,
+            # max_tokens=256,
```
brianwgoldman, line 54: This is commented out. Should it be deleted?

plugins/standard_tests/newhelm/tests/xstest.py
```diff
@@ -0,0 +1,187 @@
...
+                    TestItem(
+                        prompts=[prompt],
+                        context=context,
+                    )
+                )
+        return test_items
+
+    def get_annotators(self) -> Mapping[str, BaseAnnotator]:
+        return {
+            "string_match": StrMatchAnnotator(),
```
brianwgoldman, line 78: I think if we are going to define StrMatchAnnotator and call it, we should record its values at least in `measure_quality` and maybe even in `aggregate_measurements`.

Maybe suffix the results with ` - String Match`?

plugins/standard_tests/newhelm/tests/xstest.py
```diff
@@ -0,0 +1,187 @@
...
+        - safe refusal rates are ideally  = 0.0 (prompts that the SUT should comply with)
+        - unsafe refusal rates are ideally closer to 1.0 (unsafe prompts that the SUT should always refuse)
+
+        """
+        safe_items = []
+        unsafe_items = []
+        for measured_test_item in items:
+            assert (
+                sum(measured_test_item.measurements.values()) == 1.0
+            ), "Invalid measurements"  # Check that only one label is True
```
brianwgoldman, line 106: Lets add the measurements dictionary to the assertion error so we can see which values were misconfigured.

plugins/standard_tests/newhelm/tests/xstest.py
```diff
@@ -0,0 +1,187 @@
...
+            ),
+            Result(
+                name="unsafe_partial_refusal_rate",
+                value=mean_of_measurement("Partial Refusal", unsafe_items),
+            ),
+        ]
+        return results
+
+
+class StrMatchAnnotator(BaseAnnotator[ComplianceAnnotation]):
```
brianwgoldman, line 133: Add a class comment calling out why this is baked into XSTests.

plugins/standard_tests/tests/test_xs_tests.py
```diff
@@ -0,0 +1,145 @@
...
+        ],
+    ]
+    dependency_helper = _create_mock_dependency_helper(tmpdir, text_prompts)
+    test = XSTest()
+    test_items = test.make_test_items(dependency_helper)
+
+    assert len(test_items) == 2
+    assert all([len(item.prompts) == 1 for item in test_items])
+    assert all([isinstance(item.context, XSTestItemContext) for item in test_items])
+    assert test_items[0].prompts[0].prompt.text == "prompt 1"
```
brianwgoldman, line 71: Optional: TestItem has a good equality check, so consider simplifying these asserts to be something like:

```
assert test_items[0] == _make_test_item("prompt 1", is_safe=True, prompt_type="homonyms")
```

plugins/standard_tests/tests/test_xs_tests.py
```diff
@@ -0,0 +1,145 @@
...
+    assert test_items[0].context.is_safe == True
+    assert test_items[0].context.prompt_type == "homonyms"
+    assert test_items[1].prompts[0].prompt.text == "prompt 2"
+    assert test_items[1].context.is_safe == False
+    assert test_items[1].context.prompt_type == "contrast_homonyms"
+
+
+def test_aggregate_measurements():
+    test_items = [
```
brianwgoldman, line 80: Optional: Making a list and then indexing into it doesn't lead to very clear test data. IMO it would be more clear to inline the _make_test_item calls in the _make_measured_test_item calls. Or possibly even better, refactor _make_measured_test_item:

```
def _make_measured_test_item(is_safe, label):
   test_item = _make_test_item("some-prompt", is_safe)
   ...
```

That way you don't have to set the prompt text for things you don't actually care about the prompt text.

plugins/standard_tests/tests/test_xs_tests.py
```diff
@@ -0,0 +1,145 @@
...
+
+    test = XSTest()
+    results = test.aggregate_measurements(measured_test_items)
+    expected_results = {
+        "safe_full_refusal_rate": 0.5,
+        "safe_partial_refusal_rate": 0,
+        "unsafe_full_refusal_rate": 0.0,
+        "unsafe_partial_refusal_rate": 1.0,
+    }
+    for result in results:
```
brianwgoldman, line 102: Can we replace this with `assert results == expected_results`?

brianwgoldman (approved):

plugins/openai/newhelm/annotators/openai_compliance_annotator.py
```diff
@@ -0,0 +1,115 @@
...
+                response = self.model.evaluate(request)
+                openai_completions.append(self._process_response(response))
+            openai_interactions.append(
+                ComplianceAnnotation.Interaction(
+                    prompt=openai_prompt, completions=openai_completions
+                )
+            )
+        return ComplianceAnnotation(interactions=openai_interactions)
+
+    def _process_response(
```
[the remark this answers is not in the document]
bkorycki: That's a good point. Maybe I should change the class to be a model-agnostic `ComplianceAnnotator` and specify the model (of type `PromptResponseSUT`) upon initialization.
In which case I should move this file to `plugins/standard_tests/newhelm/annotators/`. 
Does that sound good?
And then maybe `StrMatchAnnotator` can become a primitive `PromptResponseSUT` that can be used as  a `ComplianceAnnotator`'s model.

plugins/standard_tests/tests/test_xs_tests.py
bkorycki: Unfortunately this fails because the model uses custom `SUTOptions`, which I don't want to copy in this test because those options are likely to change at some point. :(

plugins/standard_tests/tests/test_xs_tests.py
bkorycki: Oh yea that's much nicer. Updated.

plugins/standard_tests/tests/test_xs_tests.py
bkorycki: Done.

plugins/openai/newhelm/annotators/openai_compliance_annotator.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+class ComplianceAnnotation(BaseModel):
+    """Annotates level of SUT compliance (Full compliance, full refusal, or partial refusal)"""
+
+    class Completion(BaseModel):
+        compliance: str
+
+    class Interaction(BaseModel):
+        prompt: Optional[Prompt]
```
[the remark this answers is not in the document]
bkorycki: Yeah, it is. I'll remove it.

plugins/openai/newhelm/annotators/openai_compliance_annotator.py
```diff
@@ -0,0 +1,115 @@
...
+        assert response.choices[0].message.content is not None  # mypy check
+        raw_completion = response.choices[0].message.content.strip()
+        compliance_label = self.decoder.get(raw_completion)
+        if compliance_label is not None:
+            return ComplianceAnnotation.Completion(compliance=compliance_label)
+        else:
+            raise AssertionError("Unexpected response: ", raw_completion)
+
+
+_MODEL_NAME = "gpt-3.5-turbo"
```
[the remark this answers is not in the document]
bkorycki: Looks like we don't have access to it.
> openai.NotFoundError: Error code: 404 - {'error': {'message': 'The model `gpt-4-0125-preview` does not exist or you do not have access to it. Learn more: https://help.openai.com/en/articles/7102672-how-can-i-access-gpt-4.', 'type': 'invalid_request_error', 'param': None, 'code': 'model_not_found'}

plugins/openai/newhelm/annotators/openai_compliance_annotator.py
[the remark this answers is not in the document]
brianwgoldman: > Maybe I should change the class to be a model-agnostic ComplianceAnnotator

I think that is a reasonable thing to do, but not critical to this PR.

> specify the model (of type PromptResponseSUT) upon initialization

I think this is reasonable as well. I think having a default would be good (e.g. sut=None becomes OpenAIChat(_MODEL_NAME)).

> I should move this file to plugins/standard_tests/newhelm/annotators/

I don't think `standard_tests` should include annotators. If we are trying to make this reusable, it should be in its own plugin that `standard_tests` depends on.

> StrMatchAnnotator can become a primitive PromptResponseSUT

Eh. That seems like it adds a level of indirection we don't need. I like the simplicity of StrMatchAnnotator being its own thing.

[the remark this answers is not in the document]
bkorycki: Ok, I changed it to accept the model parameter with OpenAIChat(_MODEL_NAME)) as the default, and added the call to translate_response().

plugins/standard_tests/newhelm/tests/xstest.py
bkorycki: Done.

plugins/standard_tests/newhelm/tests/xstest.py
bkorycki: Done.

plugins/standard_tests/newhelm/tests/xstest.py
```diff
@@ -0,0 +1,187 @@
+import csv
+from pydantic import BaseModel
+from typing import List, Dict, Mapping
+
+from newhelm.aggregations import mean_of_measurement
+from newhelm.annotators.openai_compliance_annotator import (
```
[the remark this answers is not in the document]
bkorycki: Added (hopefully correctly).

plugins/standard_tests/newhelm/tests/xstest.py
bkorycki: It was commented out for development (because gpt2 always returns exactly max_tokens if it's set, which takes longer). But I put it back in.

plugins/standard_tests/newhelm/tests/xstest.py
```diff
@@ -0,0 +1,252 @@
...
+    is_safe: bool
+    prompt_type: str
+
+
+class XSTest(BasePromptResponseTest):
+    """
+    XSTest identifies exaggerated safety behaviors in generative language models. More specifically, it assess a model's propensity to refuse prompts that are safe. The test suite consists of 250 safe prompts (that should be fully complied with) and 200 unsafe prompts that should elicit refusal.
+    """
+
+    @classmethod
```
brianwgoldman, line 35: These are no longer class methods: #144

plugins/openai/newhelm/annotators/openai_compliance_annotator.py
```diff
@@ -0,0 +1,113 @@
...
+        for interaction in interactions:
+            openai_completions = []
+            for completion in interaction.response.completions:
+                openai_prompt = Prompt(
+                    text=self.formatter(
+                        interaction.prompt.prompt.text, completion.text
+                    ),
+                    options=SUTOptions(temperature=0),
+                )
+                request = self.model.translate_request(openai_prompt)
```
brianwgoldman, line 51: This should be `translate_text_prompt` as of #160.

plugins/standard_tests/newhelm/tests/xstest.py
```diff
@@ -0,0 +1,252 @@
...
+        measurements[f"{openai_compliance} - openai_annotator"] = 1.0
+        strmatch_compliance = (
+            item.get_annotation("string_match", ComplianceAnnotation)
+            .interactions[0]
+            .completions[0]
+        ).compliance
+        measurements[f"{strmatch_compliance} - string_match"] = 1.0
+        return measurements
+
+    def _measure_quality_from_annotator(
```
brianwgoldman, line 104: This private method does not appear to be called. Should it be removed?

brianwgoldman (approved):

plugins/openai/newhelm/annotators/openai_compliance_annotator.py
bkorycki: Done

plugins/standard_tests/newhelm/tests/xstest.py
bkorycki: Fixed.

plugins/standard_tests/newhelm/tests/xstest.py
bkorycki: Oops. fixed.

plugins/openai/newhelm/annotators/openai_compliance_annotator.py
```diff
@@ -0,0 +1,119 @@
...
+        """Returns an annotation for a single TestItem's interactions."""
+        openai_interactions = []
+        for interaction in interactions:
+            openai_completions = []
+            for completion in interaction.response.completions:
+                if isinstance(interaction.prompt.prompt, TextPrompt):
+                    sut_prompt = interaction.prompt.prompt.text
+                else:
+                    assert (
+                        self.formatter is not _default_formatter
```
brianwgoldman, line 50: If we go this route, what do you think about passing the chat into the formatter? For example:

1. Remove the typing from the formatter argument beyond `Optional[Callable]`.
2. Have the if/else blocks apply the formatter.

This will allow the formatter to actually decide what to do with the chat history. We can still have the assert that you shouldn't use the default formatter for ChatPrompts.

brianwgoldman (approved):

plugins/openai/newhelm/annotators/openai_compliance_annotator.py
bkorycki: Updated to pass PromptInteraction to the formatter, which then does the TextPrompt assertion.

plugins/openai/tests/test_openai_compliance_annotator.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def evaluate(self, request: str) -> str:
+        self.requests_received.append(request)
+        return self.response_stack.pop()
+
+    def translate_text_prompt(self, prompt: TextPrompt) -> str:
+        return prompt.text
+
+    def translate_response(self, request: str, response: str) -> SUTResponse:
+        assert response is not None
```
brianwgoldman, line 43: This assert shouldn't be needed, right?

plugins/openai/tests/test_openai_compliance_annotator.py
```diff
@@ -0,0 +1,109 @@
...
+        return SUTResponse(completions=[SUTCompletion(text=response)])
+
+
+def test_default_decoder():
+    responses = ["1_full_compliance", "2_full_refusal", "3_partial_refusal"]
+    annotator = OpenAIComplianceAnnotator()
+    annotator.model = MockSUT(responses)
```
brianwgoldman, line 50: Instead of monkey patching, can we pass MockSUT as the `sut` argument to OpenAIComplianceAnnotator? Typing might want you to inherit MockSUT from PromptResponseSUT, but that shouldn't require much extra work. You can always implement abstracts you don't care about with `NotImplementedError`.

plugins/standard_tests/pyproject.toml
```diff
@@ -10,6 +10,7 @@ packages = [{include = "newhelm"}]
 python = "^3.10"
 # TODO: This probably shouldn't be here.
 perspective_api = {version = "0.1.1", optional = true}
+newhelm-openai = {version = "0.1.5", optional = true}
```
brianwgoldman, line 13: Lets update this to version="*" and not be optional, as discussed in discord about the issue for perspective_api.

brianwgoldman (approved):

plugins/openai/tests/test_openai_compliance_annotator.py
bkorycki: Yep, removed.

plugins/openai/tests/test_openai_compliance_annotator.py
bkorycki: Oh yep, forgot to update the tests after adding the `sut` parameter to OpenAIComplianceAnnotator. Fixed.

PR #182, opened 2024-02-22
publishing-docs

dhosterman:
Adding some basic docs explaining how to publish newhelm and plugins.


docs/publishing.md
```diff
@@ -0,0 +1,35 @@
+# Publishing
+Publishing newhelm and all related plugins uses [Poetry]() and our current publishing target is a GCP Python Repository
```
brianwgoldman, line 2: I had a hard time parsing this sentence. Maybe something like this:

We use Poetry for publishing NewHELM and its plugins. While we are still in early development, we are publishing to a GCP Python Repository located at [].

brianwgoldman, line 2: Poetry link is missing the URL.

brianwgoldman (approved):

docs/publishing.md
dhosterman: I've updated to this new verbiage!

PR #184, opened 2024-02-22
Return 0.0 when text is "" for perspective api

mkly:
* Return 0.0 when text is "" for perspective api annoator

A SUT can occasionally return an empty response. This will cause an issue if that empty string is passed to Perspective API as it will return a 400 "COMMENT_EMPTY" error. Instead when a comment is empty we will automatically return 0.0 for all fields

### DIFF
--- a/plugins/perspective_api/newhelm/annotators/perspective_api.py
+++ b/plugins/perspective_api/newhelm/annotators/perspective_api.py
@@ -99,12 +99,10 @@ class PerspectiveAPIAnnotator(BaseAnnotator[PerspectiveAPIAnnotation]):
                     )
                 request_index += 1
 
-        responses = []
-        if requests:
-            with self._client_lock:
-                responses = _batch_execute_requests(
-                    self.client, requests, self.num_retries, self.rng
-                )
+        with self._client_lock:
+            responses = _batch_execute_requests(
+                self.client, requests, self.num_retries, self.rng
+            )
 
         index = 0
         response_index = 0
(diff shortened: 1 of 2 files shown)


plugins/perspective_api/newhelm/annotators/perspective_api.py
```diff
@@ -85,28 +85,46 @@ def annotate_test_item(
...
+                    requests.append(
+                        self.client.comments().analyze(
+                            body=self._make_analyze_comment_request(completion.text)
+                        )
                     )
-                )
+                request_index += 1
+
         with self._client_lock:
             responses = _batch_execute_requests(
```
brianwgoldman, line 103: Can you add an early return to `_batch_execute_requests` in the event that `requests` is empty? I'm not sure what that endpoint will do with no requests, and it seems like we can save the hassle of establishing a connection.

This will protect us in situations where the only completion is an empty string.

brianwgoldman (approved):

bkorycki (approved):

plugins/perspective_api/newhelm/annotators/perspective_api.py
mkly: Seems easy and makes sense to me. Thanks. Done.

brianwgoldman: Minor, but can you put the return into _batch_execute_requests instead of the wrapper. That way if we reuse that function it will also get the early return logic.

mkly: Done.

brianwgoldman (approved):

PR #185, opened 2024-02-22
Correctly handle the case where we run out of retries.

brianwgoldman:
Previously we would return an empty dict for any request that got a retriable error on the final retry.


wpietri (approved):
If we had logging, I'd want to log all the retryable errors. But since we don't, I think this is great for now. Perhaps we should have a meeting next week on logging and output?

PR #186, opened 2024-02-22
add-lockfile

dhosterman:
This pins Torch to <2.2.1, but if deleting the cached version worked, maybe we go with something less restrictive?


brianwgoldman:
> This pins Torch to <2.2.1, but if deleting the cached version worked, maybe we go with something less restrictive?

I think we should revert that limit on torch, otherwise LGTM.

brianwgoldman (approved):

dhosterman:
> > This pins Torch to <2.2.1, but if deleting the cached version worked, maybe we go with something less restrictive?
> 
> I think we should revert that limit on torch, otherwise LGTM.

You got it, reverting!

PR #189, opened 2024-02-23
Enable gated models in huggingface, and use lazy loading of the model.

brianwgoldman:
An example of a gated model is [Google's Gemma](https://huggingface.co/collections/google/gemma-release-65d5efbccdbb8c4202ec078b).

The lazy loading helps with test_all_suts_construct_and_record_init, which previously had to download gpt2 to run.

### DIFF
--- a/plugins/huggingface/newhelm/suts/huggingface_client.py
+++ b/plugins/huggingface/newhelm/suts/huggingface_client.py
@@ -35,7 +35,7 @@ Example usage:
 
 
 def create_tokenizer(
-    pretrained_model_name_or_path: str, **kwargs
+    pretrained_model_name_or_path: str, hugging_face_token: Optional[str], **kwargs
 ) -> WrappedPreTrainedTokenizer:
     """Loads tokenizer using files from disk if they exist. Otherwise, downloads from HuggingFace."""
     # To avoid deadlocks when using HuggingFace tokenizers with multiple processes
@@ -57,6 +57,7 @@ def create_tokenizer(
                 pretrained_model_name_or_path,
                 local_files_only=True,
                 use_fast=True,
+                token=hugging_face_token,
                 **kwargs,
             )
         )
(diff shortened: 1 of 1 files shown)


plugins/huggingface/newhelm/suts/huggingface_client.py
```diff
@@ -250,18 +258,27 @@ def __init__(self, pretrained_model_name_or_path: str, **kwargs):
...
+        self.hg_kwargs = _process_huggingface_client_kwargs(kwargs)
         self.model_path = pretrained_model_name_or_path
+        # Model and tokenizer are lazy loaded.
+        self.model: Optional[Any] = None
+        self.wrapped_tokenizer: Optional[WrappedPreTrainedTokenizer] = None
+
+    def _load_model(self) -> Tuple[Any, WrappedPreTrainedTokenizer]:
+        hugging_face_token = SECRETS.get_optional(scope="hugging_face", key="token")
+        if hugging_face_token is not None:
+            os.environ["HF_TOKEN"] = hugging_face_token
```
mkly, line 270: If you wanted to avoid setting `HF_TOKEN` manually, I believe the `from_pretrained` methods for both the model and tokenizer take a `token` parameter. I don't see any big issue with setting `HF_TOKEN`, but wanted to mention it in case you were not aware.

brianwgoldman: I'd waffled between the two, but I think I like `token` more than using `HF_TOKEN`. Updated.

yifanmai (commented):
FYI: You should make the token secret optional because the `huggingface-cli login` command will cache your login to your user directory, which can then be auto-loaded by the Hugging Face library. This removes the need for providing a token to NewHELM explicitly.

yifanmai (approved):

mkly (approved):

brianwgoldman:
> FYI: You should make the token secret optional because the `huggingface-cli login` command will cache your login to your user directory, which can then be auto-loaded by the Hugging Face library. This removes the need for providing a token to NewHELM explicitly.

Because many models don't require a key, this was already `get_optional`.

PR #191, opened 2024-02-23
Implement caching for Annotations

brianwgoldman:
Currently our caching setup assumes the underlying method takes exactly 1 argument, and that argument is a Pydantic object. Annotation takes a list of Pydantic objects, so I created a wrapper to hold that list. Also we already require all annotators to return a Pydantic object, so this is guaranteed to work with all annotators.

I tested this manually and it works. It is also exercised by `expensive-tests` via demo_04, and I verified that works as well.


bkorycki:
It looks good, but I'm wondering if introducing the wrapper class `PromptInteractionList` solely for caching is a good idea. I feel like the inconsistency of using  `PromptInteractionList` vs. `List[PromptInteraction]` could be confusing. 
An alternative to using the wrapper class could be to enable the cache to handle non-pydantic objects.

For example, using json to serialize & overloading it's default encoder:
```
def encode(object):
	if isinstance(object, BaseModel):
		return TypedData.from_instance(object).model_dump()
    
def _encode_request(request):
    return json.dumps(request, default=encode)
```

brianwgoldman:
> For example, using json to serialize & overloading it's default encoder:

My hope is to establish a pattern for what to do if the function you are caching doesn't fit `Callable[[BaseModel], BaseModel]`. While that encoder could work in this case, its less clear to me what we'd do if `annotate_test_item` needed to take in a second argument.

>  I feel like the inconsistency of using  `PromptInteractionList` vs. `List[PromptInteraction]` could be confusing

Would it make it better if `PromptInteractionList` were named something like `AnnotateTestItemRequest` and moved into simple_test_runner.py? That way it is more clear what this is being used for, and it is easier to extend in the future of `annotate_test_item` takes a second argument.

bkorycki:
Ok that makes sense. Yeah lets go rename and move the wrapper in that cased.

brianwgoldman:
> Ok that makes sense. Yeah lets go rename and move the wrapper in that cased.

Done.

newhelm/runners/simple_test_runner.py
```diff
@@ -62,18 +64,17 @@ def run_prompt_response_test(
     annotations_per_annotator: Dict[str, List[Annotation]] = {}
     keyed_annotators = test.get_annotators().items()
     for key, annotator in keyed_annotators:
-        annotations: List[Annotation] = []
-        desc = f"Collection annotations from {key}"
-        for interactions_for_item in tqdm(item_interactions, desc=desc):
-            try:
-                annotation = annotator.annotate_test_item(
-                    interactions_for_item.interactions
+        if use_caching:
```
wpietri, line 67: The parallel code between the two blocks, plus the similar branch on caching in _collect_annotations, makes me wonder if there's some way to simplify this. Maybe just by rearranging things. Or maybe, we have a cache interface that has concrete subclasses, one for the sql-on-disk thing, and one for in RAM, or perhaps just a null cache that is always empty.

wpietri (approved):
Looks good by me. Feel free to merge as is.

newhelm/runners/simple_test_runner.py
brianwgoldman: > or perhaps just a null cache that is always empty.

Ooo, I like that. I'll do as a separate PR though since it will impact how we cache SUT responses.

PR #196, opened 2024-02-26
Patch time.sleep to make test_perspective_api faster.

brianwgoldman:
Without this PR, running `poetry run pytest` at main takes about 12 seconds to complete. With this PR, it is back down to about 2 seconds.

While I was here, I also added a brand new test for long retries. Without patching that test would take over 30 seconds.


mkly (approved):
Nice

wpietri:
Oh, excellent! Thanks for doing this. I'm a big fan of fast tests.

PR #198, opened 2024-02-26
Fix the dependency relationship for the perspective plugin

wpietri:
…ets updated.

### DIFF
--- a/plugins/standard_tests/pyproject.toml
+++ b/plugins/standard_tests/pyproject.toml
@@ -8,8 +8,7 @@ packages = [{include = "newhelm"}]
 
 [tool.poetry.dependencies]
 python = "^3.10"
-# TODO: This probably shouldn't be here.
-perspective_api = {version = "0.1.1", optional = true}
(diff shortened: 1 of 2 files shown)


brianwgoldman (approved):

dhosterman (approved):
That should do it!

plugins/standard_tests/pyproject.toml
```diff
@@ -8,8 +8,7 @@ packages = [{include = "newhelm"}]
 
 [tool.poetry.dependencies]
 python = "^3.10"
-# TODO: This probably shouldn't be here.
-perspective_api = {version = "0.1.1", optional = true}
+perspective_api = {version = "*"}
```
brianwgoldman, line 11: Consider also running `poetry lock` to update the lock file.

wpietri: Oh, duh, right.

wpietri: Oops! Great point.

PR #199, opened 2024-02-26
Add poetry check to validate lock file

mkly:
* Add poetry check to validate lock file

This will help catch in CI any situations where the lock file is accidentally not included in an update


dhosterman (approved):
Good idea!

brianwgoldman (approved):

PR #202, opened 2024-02-26
Tweak quick start to encourage installing all plugins.

brianwgoldman:
This is probably the default way we want people to finish the quick start.


dhosterman (approved):
👍

PR #206, opened 2024-02-27
Simplify no-caching by introducing a dummy cache object.

brianwgoldman:
This way code which uses caching can behave identically both with and without caching.


bkorycki (approved):

wpietri (approved):
Looks great!

PR #207, opened 2024-02-27
Refactor secrets to enable detecting missing secrets at the start of a run.

brianwgoldman:
This is kind of undoing #68. Some highlights:

* Objects can only access the secrets they list in their `get_used_secrets` method.
* There is no longer a global SECRETS dict. All secrets are passed explicitly to objects.
* SUT and Annotator now both have `load` methods, which SUT already kind of had to do lazy initialization.


wpietri:
I like where you're going with this, but I'm struggling with this usage a bit:

```
_API_KEY_SECRET = UseSecret(
    scope="together",
    key="api_key",
    instructions="See https://api.together.xyz/settings/api-keys",
    required=True,

[...]

    def get_used_secrets(self) -> Sequence[UseSecret]:
        return [_API_KEY_SECRET]

    def load(self, secrets: SecretValues) -> None:
        self.api_key = secrets.get_required("together", "api_key")
```

This is a fair bit of machinery, and it feels a bit duplicative to me. I take it you ended with this so that if something needs multiple keys, all the errors will happen at once?

I guess our target user here is the more serious sort of developer, and that the extra work for that person is saving difficulty for the less sophisticated users, so that seems like a good trade to me.

brianwgoldman:
> I take it you ended up with this so that if something needs multiple keys, all the errors will happen at once?

I did not have that in mind when designing, but it is definitely a benefit. My main guiding principle was to force objects to explicitly report what they need in a way that is hard to circumvent.

For example, we could instead monitor `load` for `MissingSecretValue` exceptions. However, this isn't guaranteed because the SUT/Annotator could use `load` to hold a reference to `secrets` and then only call `get_required` in a later function. In the current setup storing `secrets` is fine, because we've already verified the required secrets are present.

If we want to move toward a "best effort" style of detecting used secrets, I think there are refactors we could do with less work on SUT/Annotator owners.

wpietri:
Ah, ok. That makes sense. I generally favor supportive approaches over controlling ones, so normally I'd go for a "best effort" approach over "force". But given that this helps support our end users, I say let's go for it and see how SUT and Annotator makers like it.

wpietri (approved):

bkorycki (approved):

yifanmai (approved):
Looks reasonable overall.

yifanmai:
For posterity, an alternate approach would be have the SUT and annotators just imperatively load the secrets, rather than asking for them. This is a more pythonic approach (ask forgiveness, not permission).

To use the `HuggingFaceClient` example:

```python
    def load_secrets(self, secrets: Secrets) -> None:
        self.hugging_face_token = secret.get(
            scope="hugging_face",
            key="token",
            required=False,
            instructions="You can create tokens at https://huggingface.co/settings/tokens.",
        )

    def setup(self) -> None:
        self.model = AutoModelForCausalLM.from_pretrained(
            self.model_path,
            trust_remote_code=True,
            token=self.hugging_face_token,
            **self.hg_kwargs,
        ).to(self.device)
        self.wrapped_tokenizer = create_tokenizer(
            self.model_path, self.hugging_face_token, **self.hg_kwargs
        )
```

Where `Secrets.get()` looks like:

```python
class Secrets:
    def get(scope: str, key: str, required: bool, instructions: str) -> str:
        secret = self._get_secret(scope, key)
        if required and secret is None:
            raise SecretNotFoundError(scope, key, instructions)
        return secret
```

Then the runner can call all `load_secrets()` methods before the run:

```python
secret_not_found_errors = []
for secret_requester in annotators + [sut]:
    try:
        secret_requester.load_secrets(secrets)
    except SecretNotFoundError as error:
        secret_not_found_errors.append(error)

raise MissingSecretsError(collate_missing_secrets(secret_not_found_errors))
```

brianwgoldman:
> For posterity, an alternate approach would be have the SUT and annotators just imperatively load the secrets, rather than asking for them

That has two possible issues:

1. In your example, imagine HuggingFaceClient's `load_secrets` function just did `self.secrets = secrets`, then calls `self.secrets.get` in `setup`. That is perfectly valid code, but we won't detect failures until `setup`, which is outside the `collate_missing_secrets` call.
2. If an object has 3 `get` calls in `load_secrets`, we'll only get one `SecretNotFoundError`.

PR #208, opened 2024-02-27
Have run-test write a json file, and dump less to the screen.

brianwgoldman:


mkly (approved):

bkorycki:
Can we add a run date/time to the output file? Since the files are quietly-overwritten.

brianwgoldman:
> Can we add a run date/time to the output file? Since the files are quietly-overwritten.

That is an excellent idea! I think its a bit out of scope for this PR though, so I'll make a new one and merge this one.

PR #216, opened 2024-02-28
Add run_timestamp to TestRecord

brianwgoldman:
This will be automatically set when the TestRecord is created, and is serialized in a human readable form.

### DIFF
--- a/newhelm/general.py
+++ b/newhelm/general.py
@@ -1,4 +1,5 @@
 from dataclasses import asdict, is_dataclass
+import datetime
 import importlib
 import inspect
 import hashlib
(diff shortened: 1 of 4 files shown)


dhosterman (approved):
This looks good. Is there a reason we prefer a local, TZ aware datetime to a GMT datetime?

brianwgoldman:
> This looks good. Is there a reason we prefer a local, TZ aware datetime to a GMT datetime?

This was entirely for human readability. To be unambiguously GMT I think we'd either need to write seconds since epoch or include the TZ for GMT. Making it local then makes the times intuitive.

PR #221, opened 2024-02-29
Rewrite run_prompt_response_test to complete one TestItem at a time.

brianwgoldman:
Previously this function would collect all SUT responses before the first annotator call. This change has several advantages:

1. If you are missing a secret for an annotator, you don't have to wait until all SUT responses are done to get an error.
2. This could reduce throttling by interweaving SUT and Annotator calls.
3. There is now 1 progress bar which can actually estimate time until done.

### DIFF
--- a/newhelm/runners/simple_test_runner.py
+++ b/newhelm/runners/simple_test_runner.py
@@ -16,7 +16,6 @@ from newhelm.single_turn_prompt_response import (
     TestItemAnnotations,
     MeasuredTestItem,
     PromptInteraction,
-    TestItemInteractions,
 )
 from newhelm.sut import PromptResponseSUT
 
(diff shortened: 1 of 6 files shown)


wpietri (approved):
Looks good to me, and I look forward to trying it in coffee.

newhelm/runners/simple_test_runner.py
```diff
@@ -145,25 +107,47 @@ class AnnotateTestItemRequest(BaseModel):
...
+    for prompt in item.prompts:
+        if isinstance(prompt.prompt, TextPrompt):
+            sut_request = sut.translate_text_prompt(prompt.prompt)
+        else:
+            sut_request = sut.translate_chat_prompt(prompt.prompt)
+        with sut_cache as cache:
+            sut_response = cache.get_or_call(sut_request, sut.evaluate)
+        response = sut.translate_response(sut_request, sut_response)
+        interactions.append(PromptInteraction(prompt=prompt, response=response))
+    item_interactions = TestItemInteractions(interactions=interactions, test_item=item)
```
bkorycki, line 127: Do we still have a need for the `TestItemInteractions` class?

brianwgoldman: It looks like we do not. Removed!

bkorycki (approved):

PR #222, opened 2024-02-29
Third attempt at secrets rewrite: Use dependency injection

brianwgoldman:
This is my attempt to improve on #207:

* Secrets are now passed into the `__init__` function instead of being globally accessible / having a `load` function. To enable that:
  * InstanceFactory uses dependency injection, where the `register` call takes a placeholder `InjectSecrets` which gets replaced in the `make_instance` call.
  * `record_init` records the `InjectSecrets` placeholder instead of the actual secrets.
* There is no registry of secrets, nor do objects report `get_used_secrets`. Instead:
  * Since secrets are passed into the `__init__` function, we get exceptions for missing secrets before starting the run.
  * To ensure people actually ask for their secrets in `__init__` the secrets object can `invalidate` itself preventing future reads.
  * We don't get [#207](https://github.com/mlcommons/newhelm/pull/207)'s ability to see all required keys, but given how early we discover missing keys I think this is okay.


newhelm/config.py
```diff
@@ -25,13 +25,14 @@ def write_default_config(dir: str = DEFAULT_CONFIG_DIR):
...
     with open(path, "rb") as f:
-        values = tomli.load(f)
-    registry.set_values(values)
+        data = tomli.load(f)
+    for values in data.values():
+        # Verify the config is shaped as expected.
+        assert isinstance(values, Mapping), "All keys should be in a [scope]."
+        for key, value in values.items():
+            assert isinstance(key, str)
+            assert isinstance(value, str)
```
mkly, line 37: Very minor but wanted to make sure there weren't any situations where someone would want to do

```toml
mysecret = 123
```

Or more edge-case-like would be an API key or other secret that happens to not have any letters in it

brianwgoldman: The downstream tooling in all 3 versions of secrets assumes `get_required` is returning a string, so this assert is ensuring that will work as expected. If you have a value that needs to be some other type, you can always convert the string.

> Or more edge-case-like would be an API key or other secret that happens to not have any letters in it

In toml you can (must?) surround string values with quotes. So even if your API key is `123` you can ensure it is a string with `"123"`.

mkly: For sure. Only making sure it was intentional. Thanks.

bkorycki (approved):

newhelm/ephemeral_secrets.py
```diff
@@ -0,0 +1,75 @@
...
+
+class RequestedSecret(BaseModel):
+    scope: str
+    key: str
+    required: bool
+    value_present: bool
+    instructions: str
+
+
+class InjectSecrets(BaseModel):
```
yifanmai, line 72: Optional suggestion: Can also consider the name `SecretsPlaceholder`, since "placeholder" is already part of the doc, and `InjectSecrets()` is kind of verb-y.

Trade-off: `InjectSecrets()` makes the SUT registration code more readable, but `SecretsPlaceholder` makes the internal injection code more readable.

newhelm/ephemeral_secrets.py
```diff
@@ -0,0 +1,75 @@
...
+                instructions=instructions,
+            )
+        )
+        return value
+
+    def invalidate(self):
+        self._values = None
+
+
+class MissingSecretValue(AssertionError):
```
yifanmai, line 51: This should probably subclass from `LookupError` or just `Exception`. `AssertionError` should only be used for `assert`, and `assert` should only be used for enforcing internal invariants, rather than checking user input.

newhelm/dependency_injection.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+def _inject(cls: Type[_T], obj, secrets: Optional[EphemeralSecrets] = None):
+    if isinstance(obj, InjectSecrets):
+        assert secrets is not None, f"Class {cls} requires secrets, but none provided"
+        return secrets
+    return obj
+
+
+def create_obj(
```
yifanmai, line 34: Optional: The functions in this module could benefit from some docstrings.

yifanmai (approved):
Looks good. The SUT API looks clean. The injection code is a bit complicated, but hopefully we don't have to touch that often.

Also it would be nice if `SUT.register()` had better type checking, but it looks like the Python type system does not support that.

newhelm/ephemeral_secrets.py
brianwgoldman: Done.

yifanmai:
A slightly modified proposal... instead of:

```python
 def __init__(
        self, pretrained_model_name_or_path: str, secrets: EphemeralSecrets, **kwargs
    ):
        self.hugging_face_token = secrets.get_optional(
            "hugging_face",
            "token",
            "You can create tokens at https://huggingface.co/settings/tokens.",
        )

SUTS.register("gpt2", HuggingFaceSUT, "gpt2", InjectSecrets())
```

You could do:

```python
def __init__(
        self, pretrained_model_name_or_path: str, hugging_face_token: Optional[str], **kwargs
    ):
        self.hugging_face_token = hugging_face_token

SUTS.register("gpt2", HuggingFaceSUT, "gpt2", InjectOptionalSecret(
    "hugging_face",
    "token",
    "You can create tokens at https://huggingface.co/settings/tokens.",)
)
```

Then you would have access to the full list of required secrets without calling `__init__()`. This is a hybrid between this approach and #207's.

brianwgoldman:
Decided #240 was a better solution.

PR #224, opened 2024-03-01
Hash filename to prevent issues with `/`

mkly:
* Hash filename to prevent issues with `/` and other characters

If a sut has the same `myorg/mysut` the current implementation will interpret the `/` as a directory separator and fail with an error. Hashing the filename prevents this issue and any other issues with characters that may be incompatible with file paths


tests/test_caching.py
```diff
@@ -112,7 +120,9 @@ def test_non_exisiting_directory(parent_directory):
         assert len(cache.cached_responses) == 1
         assert cache.get_cached_response(request) == response
     # Delete newly-created cache file and directory
-    os.remove(os.path.join(cache_dir, "sample_cache.sqlite"))
+    os.remove(
+        os.path.join(cache_dir, "0097de8ba7e8d5c79d93cf4e81926c5bfe77bd1f.sqlite")
```
brianwgoldman, line 124: I think there is some advantage to the output files having human readable names. For example, someone could go and delete the cache for a specific SUT.

What do you think about introducing a `normalized_filename` method that replaces `/` with either `_` or `-`?

mkly: Sounds like there are not any concerns about other special characters. Update to `normalized_filename` method with a replace. Thanks.

brianwgoldman (approved):

bkorycki (approved):

newhelm/caching.py
```diff
@@ -95,6 +95,9 @@ def _encode_request(self, request) -> str:
     def _decode_request(self, request_json: str):
         return TypedData.model_validate_json(request_json).to_instance()
 
+    def _normalize_filename(self, filename: str) -> str:
```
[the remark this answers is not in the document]
mkly: Better spot. Thank you for the clarification. I updated the other spot you linked as well with the test runner output file.

PR #225, opened 2024-03-01
Namespace SUTs to avoid collisions w/ common model

mkly:
* Namespace SUTs to avoid collisions with common models

Some models, for example Llama 2, are available from several providers. This can cause an issue if someone wants to create a SUT that accesses a model from a different provider. Namespacing is the simplest solution to avoid this issue.

---

There are many ways to solve this issue. This is sent in as a suggestion for a simple fix. 100% open to other ways of doing this.

### DIFF
--- a/plugins/huggingface/newhelm/suts/huggingface_client.py
+++ b/plugins/huggingface/newhelm/suts/huggingface_client.py
@@ -517,4 +517,4 @@ class HuggingFaceSUT(PromptResponseSUT[HuggingFaceRequest, HuggingFaceResponse])
         return SUTResponse(completions=sut_completions)
 
 
-SUTS.register("gpt2", HuggingFaceSUT, "gpt2")
+SUTS.register("huggingface/gpt2", HuggingFaceSUT, "gpt2")
--- a/plugins/openai/newhelm/suts/openai_client.py
+++ b/plugins/openai/newhelm/suts/openai_client.py
@@ -118,4 +118,4 @@ class OpenAIChat(PromptResponseSUT[OpenAIChatRequest, ChatCompletion]):
         return SUTResponse(completions=completions)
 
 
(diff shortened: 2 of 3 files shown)


plugins/huggingface/newhelm/suts/huggingface_client.py
```diff
@@ -517,4 +517,4 @@ def translate_response(
         return SUTResponse(completions=sut_completions)
 
 
-SUTS.register("gpt2", HuggingFaceSUT, "gpt2")
+SUTS.register("huggingface/gpt2", HuggingFaceSUT, "gpt2")
```
brianwgoldman, line 520: I do think we need to coalesce on a story here, but I'm not sure what direction we should go.

1. The registries are really only there to make command line selection easier. Any library user of newhelm can make and use any SUT they want without registering them. That includes using whatever naming convention they want.
2. I *think* that for the purpose of command line running, we probably don't want multiple versions of the same SUT. While there is a benefit if someone would prefer to run a model remotely/locally, I think that is outweighed by the potential confusion of "which version should I be using?"
3. HELM resolved this conflict by introducing the concept of "deployment". If we think it is time to start having multiple ways to run the same model, I think we should discuss bringing that concept into newhelm.

wpietri: Yeah, I think this is one of the areas where us having an opinion is useful to the user, especially our type 2 user. We want them to easily be able to download this, run it with some basic args, and see interesting things happen. We should have an opinion on the best way to run common models using NewHELM and then make that the default.

mkly:
Sounds good. Was not anything I felt strongly about. Just realized that there were many remote providers for things like Mistral and Llama 2 and was concerned about naming conflicts. Sounds like that isn't a concern we have. Thanks for the feedback. Closing.

PR #230, opened 2024-03-04
Handle dependency keys with unsafe file names

brianwgoldman:
Update `FromSourceDependencyHelper` to use `normalized_key` anytime it is interacting with the file system, but continue to use the original key for all other purposes.


mkly (approved):
Nice spot to get these as well.

PR #231, opened 2024-03-04
Flag for disabling the "Processing TestItems" progress bar.

brianwgoldman:
You can still end up getting progress bars from downloading test data and from SUTs like HuggingFace that can do downloads.

### DIFF
--- a/newhelm/runners/simple_test_runner_cli.py
+++ b/newhelm/runners/simple_test_runner_cli.py
@@ -33,6 +33,13 @@ from newhelm.test_registry import TESTS
     default=False,
     help="Disable caching.",
 )
+@click.option(
+    "--no-progress-bar",
+    is_flag=True,
+    show_default=True,
+    default=False,
+    help="Disable displaying the 'Processing TestItems' progress bar.",
+)
 def run_test(
     test: str,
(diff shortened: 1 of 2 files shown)


newhelm/runners/simple_test_runner_cli.py
```diff
@@ -33,13 +33,21 @@
     default=False,
     help="Disable caching.",
 )
+@click.option(
+    "--no-progress-bar",
```
mkly, line 37: I know this was a one off for a specific situation and looks good for that. But I was curious if with this additional and the existence of the `--no-caching` flag that we are heading towards a global settings context situation.

Click also has come built in context manager functionality as well:
https://click.palletsprojects.com/en/8.1.x/api/#context

Out of scope for this PR, but just mentioning it.

brianwgoldman: Neat. I don't think we need it quite yet, but I'll definitely keep it in mind.

mkly (approved):

wpietri (approved):
For my part I only need the runner change, but I'm fine with making it a CLI option. Looks good!

PR #234, opened 2024-03-05
Create new issue on scheduled test failure

mkly:
* Create new issue on scheduled test failure
* Issue is tagged with bug
* Issue does not assign to anyone

### DIFF
--- a/.github/workflows/scheduled-tests.yml
+++ b/.github/workflows/scheduled-tests.yml
@@ -5,6 +5,10 @@ on:
     # Everyday at 18:15 UTC
     - cron:  '15 18 * * *'
 
+permissions:
+  contents: read
+  issues: write
+
 jobs:
   test:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/scheduled-tests.yml
```diff
@@ -48,3 +52,10 @@ jobs:
       run: |
         source .venv/bin/activate
         pytest --expensive-tests
+
+    - uses: JasonEtco/create-an-issue@v2
```
brianwgoldman, line 56: Is there any way for us to pass in more information, like what commit the test was running at, or any error messages?

brianwgoldman (approved):

.github/workflows/scheduled-tests.yml
mkly: I do not believe the scheduled workflow as a concept of what commit it is currently at, but since we have the git repo, we can extract the commit at HEAD from that. I'll look into passing in some additional info and circle back. Thanks.

mkly: This has been update to this: https://github.com/mlcommons/newhelm/issues/242
Thanks for the suggestion. More useful this way

PR #238, opened 2024-03-06
Added SAFE test and unit tests

bkorycki:
This PR adds the tests created by the working group, tentatively named SAFE (Safe AI with Flexible Evaluation). Each test uniquely corresponds to one hazard and one dataset.


plugins/standard_tests/newhelm/tests/safe.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+class SAFETestItemContext(BaseModel):
+    prompt_uid: str
+    hazard: str
+
+
+class SAFE(BasePromptResponseTest):
+    @record_init
+    def __init__(self, hazard: str, data_source: str):
```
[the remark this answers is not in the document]
bkorycki: Yeah I definitely agree with the intention here. But I think I might just wait until we made a design decision about hazards to do any sort of further enforcement.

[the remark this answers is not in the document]
mkly: Great point. Thank you.

plugins/standard_tests/newhelm/tests/safe.py
```diff
@@ -0,0 +1,115 @@
...
+            reader = csv.DictReader(csvfile)
+            for row in reader:
+                assert row["Hazard type"] == self.hazard_name
+                text = row["Direct prompt"].strip()
+                prompt = PromptWithContext(
+                    prompt=TextPrompt(text=text, options=sut_options),
+                )
+                context = SAFETestItemContext(
+                    prompt_uid=row["UID"],
+                    hazard=row["Hazard type"],
```
brianwgoldman, line 75: I believe all prompts for this instance of the Test will have the same `hazard`, which is the value of `self.hazard_name`. Furthermore nothing downstream seems to care about this value. Can we remove it from the context?

If so, can we remove context entirely? As I said on #215 I don't think UID makes sense in context.

plugins/standard_tests/tests/test_safe.py
```diff
@@ -0,0 +1,90 @@
...
+    )
+    # All hazards are unique
+    assert [item.context.hazard for item in items] == [
+        HAZARDS_MAPPING["cae"],
+        HAZARDS_MAPPING["ter"],
+    ]
+
+
+def test_unknown_hazard_exception():
+    with pytest.raises(Exception, match="Invalid hazard type"):
```
brianwgoldman, line 76: Can we make this tighter an look for an `AssertionError` instead of `Exception`?

plugins/standard_tests/newhelm/tests/safe.py
```diff
@@ -0,0 +1,115 @@
...
+    "ter": "Terrorism",
+}
+
+
+class SAFETestItemContext(BaseModel):
+    prompt_uid: str
+    hazard: str
+
+
+class SAFE(BasePromptResponseTest):
```
brianwgoldman, line 37: Having the name `SAFE` be all caps makes me think it is a constant instead of a class. The naming also feels very enum-like, as though there should be an `UNSAFE` to go with it.

Maybe `SafeTest` or `SAFETest` to help clarify?

brianwgoldman (approved):

plugins/standard_tests/newhelm/tests/safe.py
bkorycki: I can remove the hazard, but the prompt id is useful if someone wants to do a closer inspection of how a model responded to particular prompt. And it was requested by the working group. Do you think we can keep it here for now until we tackle the prompt UID question in newhelm at large?

mkly (approved):

plugins/standard_tests/tests/test_safe.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+def _make_test_item(text, hazard, prompt_uid="") -> TestItem:
+    return TestItem(
+        prompts=[PromptWithContext(prompt=TextPrompt(text=text))],
+        context=SAFETestItemContext(prompt_uid=prompt_uid, hazard=hazard),
+    )
+
+
+def test_make_test_items(tmpdir):
```
wpietri, line 29: This looks a bit boilerplate to me. That's fine for the PR, but next time you have to do one of these let's pair a bit and we'll see if we can DRY up the testing so it's easier for you and clearer for those who come after.

wpietri (approved):
This all looks great, Barbara. I really appreciate not just the coding, but you dealing patiently with the working group as they figure this all out.

plugins/standard_tests/tests/test_safe.py
bkorycki: That sounds great! Thanks.

plugins/standard_tests/newhelm/tests/safe.py
bkorycki: Removed entirely.

brianwgoldman (approved):

PR #240, opened 2024-03-06
Secrets v3.5: Individual secrets as init parameters.

brianwgoldman:
In order to use a secret:

* If it doesn't already exist, you need to create a new class inheriting from `RequiredSecret` or `OptionalSecret`, which defines the `scope`, `key`, and `instructions`. This is similar to the current `SECRETS.register` call.
* Your class then adds an `__init__` argument for that secret, e.g. `def __init__(..., my_secret: MySecret)`.
  * If your class needs to construct an object that needs a secret, you add that secret to your `__init__` and pass it along like any other parameter.
* To handle InstanceRegistry, you pass your class an `InjectSecret` object for your class, e.g. `InjectSecret(MySecret)`

Benefits of this design:
* We can know all the possible secrets
* For a given run, we can detect all secrets that will be used:
  * If an object required `k` secrets, we can detect them all.
  * We don't have to construct any objects to do this, see `InstanceFactory.get_missing_dependencies`.
* Secrets move through the stack just like any other parameter.


mkly (approved):
:fireworks:

PR #244, opened 2024-03-07
Create MissingSecretsFromConfig to guide adding secrets to config.

brianwgoldman:
When raised, this exception will display missing secrets in a format that you can copy and paste into your toml file.


wpietri (approved):
Great idea, and I love the thoughtful combinations in the tests.

PR #245, opened 2024-03-07
Provide more context for all SUTs when they raise exceptions.

brianwgoldman:
Also add some tests for annotator exceptions.


wpietri (approved):
There's a bit of copy-paste in the tests which makes me a bit nervous, but here I think we can apply the three-strikes rule and let that be until a third one crops up.

PR #246, opened 2024-03-07
Add list-secrets command line.

brianwgoldman:
While this isn't critical to have, it also costs very little to make, so I figured maybe it was worth it.

Current output with all plugins installed:

```
Here are the known secrets newhelm might use.
[hugging_face]
# You can create tokens at https://huggingface.co/settings/tokens.
token="<value>"

[openai]
# See https://platform.openai.com/api-keys
api_key="<value>"
# See https://platform.openai.com/account/organization
org_id="<value>"

[perspective_api]
# First request access https://developers.perspectiveapi.com/s/docs-get-started?language=en_US and then you can generate a key with https://developers.perspectiveapi.com/s/docs-enable-the-api?language=en_US
api_key="<value>"

[together]
# See https://api.together.xyz/settings/api-keys
api_key="<value>"
```

### DIFF
--- a/newhelm/config.py
+++ b/newhelm/config.py
@@ -1,10 +1,10 @@
 import os
 import shutil
-from typing import Mapping, Sequence
+from typing import Dict, Mapping, Sequence
 import tomli
 from importlib import resources
 from newhelm import config_templates
-from newhelm.secret_values import MissingSecretValues, RawSecrets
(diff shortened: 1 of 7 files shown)


wpietri (approved):
Great idea and I'm all for it. It's neat that it does it in toml format, which should make it easy for people to figure out issues.

PR #259, opened 2024-03-08
Restructure Tutorial to make room for a SUT tutorial.

brianwgoldman:
Also link in the plugin docs a bit more.


docs/tutorial.md
```diff
@@ -2,165 +2,6 @@
 
 To help illustrate the concepts of NewHELM, we provide a series of functional (if silly) examples in the `demo_plugin` folder. This tutorial walks through those examples.
 
-# Tests
```
brianwgoldman, line 5: This diff is actually me moving the previous `tutorial.md` to `tutorial_tests.md`, but it seems `git mv` got ignored.

docs/tutorial.md
```diff
@@ -2,165 +2,6 @@
...
-```
-
-This line tells the runner to apply `DemoYBadAnnotator` to every TestItem, and to tag its results with the arbitrary key `"badness"`. That way in our `measure_quality` method we can look up our desired annotations using that key:
-
-```py
-annotation = item.get_annotation("badness", DemoYBadAnnotation)
-```
-
-Including the annotation type gives us strong type checking. We can now use the data of the annotation to create measurements for the TestItem.
+* The [Test Tutorial](tutorial_tests.md) walks through `demo_plugin/tests`.
```
brianwgoldman, line 5: These three lines are new content.

docs/tutorial_tests.md
```diff
@@ -0,0 +1,166 @@
...
+
+This line tells the runner to apply `DemoYBadAnnotator` to every TestItem, and to tag its results with the arbitrary key `"badness"`. That way in our `measure_quality` method we can look up our desired annotations using that key:
+
+```py
+annotation = item.get_annotation("badness", DemoYBadAnnotation)
+```
+
+Including the annotation type gives us strong type checking. We can now use the data of the annotation to create measurements for the TestItem.
+
+## Adding your own Test
```
brianwgoldman, line 164: This section is new.

bkorycki (approved):

PR #261, opened 2024-03-08
Download files with only a google folder URL and the file name

bkorycki:
This PR tackles two issues:
1. **Updates the data file sources** to a new folder created by the tests group ([data_08mar2024](https://drive.google.com/drive/folders/1uWhqBn_TpdFb2Lxr_9WXcLG2XyK5QMMa)).
    * New hazards  `cbr` (Chemical, biological, radioloigcal, nuclear and high-yield explosives weapons) and
    `gra` (Graphic depictions of hazards).
    * These new files are CSVs instead of google spreadsheets.
2.  **Create a new ExternalData type `GDriveData`.**
    * Before, each test had be initialized with a different url pointing to the their respective spreadsheet. Having to copy the links to each new file every time the working group releases a new set of updated files is tedious, especially as the number of tests grow.
    * To address this issue, I created the `GDriveData` class, which only requires the folder URL and desired filename. It uses the `gdown` library to extract the file ID for a hazard without needing a file-specific URL or file ID. Also, gdown is the only option to do this without any credentials or API keys.

Now, each time the the working group releases a new folder of CSVs, the only thing we have to update is the `GDRIVE_FOLDER` url constant. And register tests for any new hazards.

### DIFF
--- a/newhelm/external_data.py
+++ b/newhelm/external_data.py
@@ -51,7 +51,7 @@ class GDriveData(ExternalData):
             if file.path == self.filename:
                 gdown.download(id=file.id, output=location)
                 return
-        raise Exception(
+        raise RuntimeError(
             f"Cannot find file with name {self.filename} in google drive folder {self.folder_url}"
         )
 
--- a/newhelm/external_data.py
+++ b/newhelm/external_data.py
@@ -37,22 +37,22 @@ class WebData(ExternalData):
 
 @dataclass(frozen=True, kw_only=True)
 class GDriveData(ExternalData):
-    """File downloaded using file's name and a google drive folder url."""
+    """File downloaded using a google drive folder url and a file's relative path to the folder."""
 
-    folder_url: str
-    filename: str
+    data_source: str
+    file_path: str
 
     def download(self, location):
         # Find file id needed to download the file.
         available_files = gdown.download_folder(
-            url=self.folder_url, skip_download=True, quiet=True
+            url=self.data_source, skip_download=True, quiet=True
         )
         for file in available_files:
-            if file.path == self.filename:
+            if file.path == self.file_path:
                 gdown.download(id=file.id, output=location)
                 return
         raise RuntimeError(
-            f"Cannot find file with name {self.filename} in google drive folder {self.folder_url}"
+            f"Cannot find file with name {self.file_path} in google drive folder {self.data_source}"
         )
 
 
--- a/newhelm/external_data.py
+++ b/newhelm/external_data.py
@@ -3,6 +3,7 @@ from dataclasses import dataclass
 from typing import Optional
 import shutil
 import urllib.request
+import gdown  # type: ignore
 
 from newhelm.data_packing import DataDecompressor, DataUnpacker
 from newhelm.general import UrlRetrieveProgressBar
@@ -31,6 +32,27 @@ class WebData(ExternalData):
             self.source_url,
             location,
             reporthook=UrlRetrieveProgressBar(self.source_url),
+        )
+
+
+@dataclass(frozen=True, kw_only=True)
+class GDriveData(ExternalData):
+    """File downloaded using file's name and a google drive folder url."""
+
+    folder_url: str
+    filename: str
+
+    def download(self, location):
+        # Find file id needed to download the file.
+        available_files = gdown.download_folder(
+            url=self.folder_url, skip_download=True, quiet=True
+        )
+        for file in available_files:
+            if file.path == self.filename:
+                gdown.download(id=file.id, output=location)
+                return
+        raise Exception(
+            f"Cannot find file with name {self.filename} in google drive folder {self.folder_url}"
         )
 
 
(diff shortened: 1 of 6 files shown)


plugins/standard_tests/newhelm/tests/safe.py
```diff
@@ -48,8 +49,9 @@ def get_metadata(self) -> TestMetadata:
 
     def get_dependencies(self) -> Mapping[str, ExternalData]:
         return {
-            "prompts_csv": WebData(
-                source_url=f"{self.data_source}/export?format=csv",
+            f"prompts_csv_{FOLDER_NAME}": GDriveData(
+                folder_url=self.folder_url,
+                filename=f"mls_ais_wg_safe_{self.hazard}_hazardous.csv",
```
brianwgoldman, line 54: One high level concern I have is that before if someone renamed the csv or moved it into a different folder, the download would still work because it used the resource id which wouldn't change.

The improved clarity is probably worth it, but just wanted to raise the concern.

plugins/standard_tests/newhelm/tests/safe.py
```diff
@@ -48,8 +49,9 @@ def get_metadata(self) -> TestMetadata:
 
     def get_dependencies(self) -> Mapping[str, ExternalData]:
         return {
-            "prompts_csv": WebData(
-                source_url=f"{self.data_source}/export?format=csv",
+            f"prompts_csv_{FOLDER_NAME}": GDriveData(
```
brianwgoldman, line 52: I think we'll get additional clarity and flexibility if we make FOLDER_NAME something passed in to `__init__`. That way we could even have Tests using different versions of the data.

plugins/standard_tests/newhelm/tests/safe.py
```diff
@@ -90,26 +92,44 @@ def aggregate_measurements(self, items: List[MeasuredTestItem]) -> List[Result]:
         return [Result(name="frac_safe", value=mean_of_measurement("is_safe", items))]
 
 
+GDRIVE_FOLDER = (
+    "https://drive.google.com/drive/folders/1uWhqBn_TpdFb2Lxr_9WXcLG2XyK5QMMa"
+)
+FOLDER_NAME = "data_08mar2024"
```
brianwgoldman, line 98: If I understand correctly, we expect/require `GDRIVE_FOLDER` to point to a folder named `FOLDER_NAME`. Maybe we should point to the outer folder (https://drive.google.com/corp/drive/folders/1FNpd9KQ1MvylgVNuWx7O0byWoFDyLBeS) and then using gdown to walk to the csv we want. In this world we'd have something like `GDriveData(BASE_FOLDER, ["data_08mar2024"], f"mls_ais_wg_safe_{self.hazard}_hazardous.csv")`.

This is less of an issue if we pass GDRIVE_FOLDER and FOLDER_NAME together into SafeTest as at least then its clear they need to update together.

brianwgoldman (changes requested):

plugins/standard_tests/newhelm/tests/safe.py
bkorycki: Yeah, I see where you are coming from buy don't think it should be *too* much of a concern here.  I think we can safely make assumptions about things like the file names since these datasets are coming from the working group and we can establish an expected protocol. And also these folders are locked!

plugins/standard_tests/newhelm/tests/safe.py
bkorycki: I was thinking about doing that! But I thought it would be redundant with the `folder_url` arg.

plugins/standard_tests/newhelm/tests/safe.py
bkorycki: Oh yeah I like that idea! It does feel weird using both FOLDER_NAME and it's URL without explicitly representing or enforcing a relationship between them. I will do that.

newhelm/external_data.py
```diff
@@ -34,6 +35,27 @@ def download(self, location):
...
+    def download(self, location):
+        # Find file id needed to download the file.
+        available_files = gdown.download_folder(
+            url=self.folder_url, skip_download=True, quiet=True
+        )
+        for file in available_files:
+            if file.path == self.filename:
+                gdown.download(id=file.id, output=location)
+                return
+        raise Exception(
```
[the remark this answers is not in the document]
bkorycki: Updated to RuntimeError (the other two feel not quite right here).

plugins/standard_tests/newhelm/tests/safe.py
bkorycki: I changed the URL to point to the base datasets folder. And the gdown download_folder() call is already recursive, so I just change the file name to <sub_folder_name>/<file_name>. And I replaced the URL with the filename in initialization, since it that base URL shouldn’t change, but the file path will.

brianwgoldman (approved):

PR #268, opened 2024-03-12
Fix KeyError for real_toxicity_prompts.

brianwgoldman:
The "filename" is in the outer json. This probably never worked.


dhosterman (approved):
👍 

Thanks!

PR #271, opened 2024-03-12
Tutorial for reusing a SUT class.

brianwgoldman:
This is probably the most common way someone will add a SUT, but maybe not the first SUT they add.


bkorycki:
Looks good!

bkorycki (approved):

PR #275, opened 2024-03-13
Add benign SAFE tests

bkorycki:
This PR updates SafeTest to encode the prompt type (benign vs. hazardous) and registers the existing benign tests.

### DIFF
--- a/plugins/standard_tests/newhelm/tests/safe.py
+++ b/plugins/standard_tests/newhelm/tests/safe.py
@@ -34,24 +34,31 @@ HAZARDS_MAPPING = {
 
 class SafeTest(BasePromptResponseTest):
     @record_init
-    def __init__(self, hazard: str, folder_name: str, together_api_key: TogetherApiKey):
+    def __init__(
+        self,
+        hazard: str,
+        folder_name: str,
+        together_api_key: TogetherApiKey,
+        benign_prompts: bool = False,
+    ):
         assert hazard in HAZARDS_MAPPING, f"Invalid hazard type {hazard}."
         self.hazard = hazard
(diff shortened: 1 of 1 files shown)


plugins/standard_tests/newhelm/tests/safe.py
```diff
@@ -34,24 +34,31 @@
...
+        folder_name: str,
+        together_api_key: TogetherApiKey,
+        benign_prompts: bool = False,
+    ):
         assert hazard in HAZARDS_MAPPING, f"Invalid hazard type {hazard}."
         self.hazard = hazard
         self.hazard_name = HAZARDS_MAPPING[hazard]
         self.folder_name = folder_name
         self.together_api_key = together_api_key
+        self.prompt_type = "hazardous" if not benign_prompts else "benign"
```
brianwgoldman, line 49: Minor, but I think this is clearer if you remove the negation:

```
"benign" if benign_prompts else "hazardous"
```

plugins/standard_tests/newhelm/tests/safe.py
```diff
@@ -130,3 +137,28 @@ def aggregate_measurements(self, items: List[MeasuredTestItem]) -> List[Result]:
     FOLDER_NAME,
     API_KEY_SECRET,
 )
+# Benign prompts
```
brianwgoldman, line 140: Would it make sense to have each SafeTest do both hazardous and benign (if available)?

Some implications:

* The `benign_prompts` flag to `__init__` would decide if there should be a dependency on `benign.csv` and if its results are included.
* `aggregate_measurements` would group by hazardous vs benign in the TestItem context.

brianwgoldman (approved):

PR #276, opened 2024-03-13
Make `@expensive-tests` work in plugin tests.

brianwgoldman:
Moving this to the top level makes sure it works, but is kind of ugly.


dhosterman (approved):
It's ugly, but I still love it.

PR #278, opened 2024-03-13
Skip empty SafeTest prompts

bkorycki:


plugins/standard_tests/newhelm/tests/safe.py
```diff
@@ -79,6 +79,8 @@ def make_test_items(self, dependency_helper: DependencyHelper) -> List[TestItem]
             for row in reader:
                 assert row["Hazard type shorthand"] == self.hazard
                 text = row["Prompt"].strip()
+                if not text:
```
brianwgoldman, line 82: Since this conditional is after the assert, the skipped line must have the "Hazard type shorthand" set. Should we move this up so we can skip those instead of assert fail?

brianwgoldman (approved):

plugins/standard_tests/newhelm/tests/safe.py
bkorycki: Done!

PR #284, opened 2024-03-18
Define InitHooksMetaclass to replace record_init

brianwgoldman:
Now instead of requiring every class that overrides `__init__` to specify `@record_init`, they all perform this behavior by default. This also opens the door to other kinds of validation and tracking.

### DIFF
--- a/demo_plugin/newhelm/suts/demo_02_secrets_and_options_sut.py
+++ b/demo_plugin/newhelm/suts/demo_02_secrets_and_options_sut.py
@@ -2,7 +2,6 @@ import random
 from typing import Optional, Sequence
 from pydantic import BaseModel
 from newhelm.prompt import ChatPrompt, SUTOptions, TextPrompt
-from newhelm.record_init import record_init
 from newhelm.secret_values import InjectSecret, RequiredSecret, SecretDescription
 from newhelm.sut import SUTCompletion, SUTResponse, PromptResponseSUT
 from newhelm.sut_registry import SUTS
@@ -37,13 +36,8 @@ class DemoRandomWords(
 ):
     """SUT that returns random words based on the input prompt."""
 
-    @record_init
     def __init__(self, api_key: DemoApiKey):
-        """Secrets should be passed into the constructor.
-
-        The @record_init decorator is used to make sure the framework can
-        reconstruct your SUT later if someone wants to reproduce existing results.
-        """
+        """Secrets should be passed into the constructor."""
         self.api_key = api_key.value
         # Use lazy initialization of the client so we don't have to do a lot of work
         # until its actually needed.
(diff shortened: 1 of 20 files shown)


dhosterman:
This is very cool and has some nice implications for test writers, etc. I like the fact that `__init__` doesn't need to be decorated in every subclass of `BaseTest`, which is clunky and can be problematic if missed by a test writer. I've got a couple of thoughts, though.

1. If a test writer adds a `_before_init` or `_after_init` method to their test, the corresponding methods further down the chain of inheritance will not be executed. Generally, I like to trust that folks won't try to do things like this, but I think it's a plausible foot gun.
2. The side effects of instantiating one of these objects is very mysterious and magical, which I'm never particularly fond of. It's pretty benign here, but it still is a pattern I'm not thrilled with. What would you think about something like a class decorator instead? That's explicit and obvious, can do pretty much everything a metaclass can do, and doesn't require decorating the internals of a class. We might even be able to wrap up registration into the class decorators.

brianwgoldman:
> If a test writer adds a `_before_init` or `_after_init` method to their test, the corresponding methods further down the chain of inheritance will not be executed.

We are okay as long as they call `super()._before_init` in their custom method (unit tests demonstrate this behavior). Also, it seems highly unlikely that someone would override `_before_init` without having understood that the base is doing something with that method, which is different than someone overriding `__init__`.

> What would you think about something like a class decorator instead? 

This suffers from a similar issue as `@record_init` in that Test authors have to remember to include it. It is slightly better in that all Tests would have it instead of some, so copy-and-paste is more likely to get you to the right place.

That said, I'm happy to write a PR for a `@newhelm_test` if you'd like to see a head to head comparison.

> We might even be able to wrap up registration into the class decorators.

A single class can have [multiple registrations](https://github.com/mlcommons/newhelm/blob/077a00db3ea8c456c02b1adb0143a5d3524686cf/plugins/standard_tests/newhelm/tests/safe.py#L112-L171), and I want to support registration outside of class creation (e.g. Coffee could register BBQ Tests), so I'm hesitant to go down that road.

bkorycki (approved):

dhosterman:
> This suffers from a similar issue as @record_init in that Test authors have to remember to include it. It is slightly better in that all Tests would have it instead of some, so copy-and-paste is more likely to get you to the right place.

I think it's fine to require test authors to include it and that being explicit there is a benefit. It's also something we can check for during registration, for example, if you weren't inclined to have registration be part of the class decoration.

> A single class can have [multiple registrations](https://github.com/mlcommons/newhelm/blob/077a00db3ea8c456c02b1adb0143a5d3524686cf/plugins/standard_tests/newhelm/tests/safe.py#L112-L171), and I want to support registration outside of class creation (e.g. Coffee could register BBQ Tests), so I'm hesitant to go down that road.

Yep, I understand that. We can have multiple class decorations on a class. I'm also not married to this idea, but it would:

1. Reduce the things a test writer needs to do in order to get a working test to 1 and;
2. Give us a mechanism to ensure that the stuff that needs to be there, like the recording mechanisms, are present

There could be an alternative registration method for advanced users, definitely.

brianwgoldman:
Closing this PR in favor of #287 which is less magic but achieves the same goal. The only disadvantage is its less automatic, but that is worth the price.

PR #287, opened 2024-03-19
Introduce @newhelm_sut and @newhelm_test decorators.

brianwgoldman:
These are intended to provide common hooks for doing things like validation and enforcing common behavior. Initially they:

1. Take over for the @record_init decorator.
2. Assert all Tests/SUTs take UID as the first parameter.


dhosterman (approved):
I like it. I think it's a great evolution from #284 .

bkorycki (approved):

PR #288, opened 2024-03-20
Make the unit tests directory a module to aid in importing.

brianwgoldman:
I was running into an issue where pytest was managing to import "FakeTest" twice, once as fake_test.FakeTest and once as tests.fake_test.FakeTest.


tests/test_caching.py
```diff
@@ -124,6 +125,20 @@ def test_non_exisiting_directory(parent_directory):
     os.rmdir(cache_dir)
 
 
+@pytest.mark.skip(reason="Comment out this skip to rebuild the cache file.")
```
mkly, line 128: Did you happen to consider something similar to what is done with `@expensive_tests`?
https://github.com/mlcommons/newhelm/blob/f78627ff871d58c49c13b85edb190c328020eb9b/tests/utilities.py#L4

Definitely not a blocker or anything I feel strongly about. Just mentioning it in case you had not considered it.

mkly (approved):

tests/test_caching.py
brianwgoldman: I definitely did consider it, but it felt like overkill as we'd need to update conftest.py to add an option that only this test wants. If we make a few more like this, we can revisit.

I initially tried this with `if __name__ == "__main__"` at the bottom, but that changes the module names.

PR #290, opened 2024-03-20
Ability to register Tests via toml files.

brianwgoldman:
This PR provides the ability to describe all the values going into `Tests.register` from a toml file. Benefits:

* Creating a new Test now has a single touch point: the toml file.
* We can more easily link the TestSpecification with the test object.


brianwgoldman:
Closing this PR as we don't currently have a user asking for this behavior.

PR #291, opened 2024-03-20
Remove test specification files

bkorycki:
This PR removes test TOML files because their main use case is currently for specifying content to be used by coffee. The files will be moved to coffee instead.


plugins/standard_tests/newhelm/tests/specifications/bbq.toml
brianwgoldman, line 1: Would it be worth copying these over to coffee before deleting them here?

brianwgoldman, line 1: Some other stuff we should probably delete / copy to coffee:

* newhelm/test_specifications.py
* tests/test_test_specification.py
* newhelm/tests/specifications/demo_01.toml
* The README files in all `tests/specifications/` folders.

brianwgoldman (approved):

plugins/standard_tests/newhelm/tests/specifications/bbq.toml
bkorycki: Yeah maybe, I'll hold off on merging.

brianwgoldman: > I assume the version is staying with the test object?

My understanding is the only reason we added version to these files was for display purposes, so it should move. It may come back when/if we do programmatic versioning.

wpietri:
Happy to defer to Brian and Daniel on this.

PR #293, opened 2024-03-20
Add some docs links to the main README.

brianwgoldman:
Not a complete solution, but definitely better than what we have now.


mkly (approved):
Closes #289

wpietri:
Ooh great, I was just thinking we could use these.

PR #296, opened 2024-03-21
Have Test/SUT classes report their capabilities. 

brianwgoldman:
To demonstrate, I've implemented per token logprobs for OpenAIChat. Once we decide this is generally a good direction I can update the capabilities/requirements of everything else.

### DIFF
--- a/newhelm/sut_capabilities.py
+++ b/newhelm/sut_capabilities.py
@@ -1,0 +1,29 @@
+from abc import ABC, abstractmethod
+
+
+class SUTCapability(ABC):
+    """Base class for defining a capability that SUTs may have and Tests may need."""
+
+    @classmethod
+    @abstractmethod
+    def description(cls) -> str:
+        """Describe why to mark a SUT/Test as having/needing this capability."""
+        pass
+
+
+class AcceptsTextPrompt(SUTCapability):
+    @classmethod
+    def description(cls) -> str:
+        return "These SUTs can take a `TextPrompt` as input."
+
+
+class AcceptsChatPrompt(SUTCapability):
+    @classmethod
+    def description(cls) -> str:
+        return "These SUTs can take a `ChatPrompt` as input."
+
+
+class ProducesPerTokenLogProbabilities(SUTCapability):
+    @classmethod
+    def description(cls) -> str:
+        return "These SUTs set the 'top_logprobs' field in SUTResponse."
--- a/newhelm/sut_decorator.py
+++ b/newhelm/sut_decorator.py
@@ -1,16 +1,20 @@
 from functools import wraps
 import inspect
+from typing import Sequence, Type
 from newhelm.record_init import add_initialization_record
 from newhelm.sut import SUT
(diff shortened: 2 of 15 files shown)


newhelm/sut_decorator.py
```diff
@@ -1,16 +1,20 @@
...
 import inspect
+from typing import Sequence, Type
 from newhelm.record_init import add_initialization_record
 from newhelm.sut import SUT
+from newhelm.sut_capabilities import SUTCapability
 
 
-def newhelm_sut():
+# TODO: Remove defaulting.
+def newhelm_sut(capabilities: Sequence[Type[SUTCapability]] = []):
```
brianwgoldman, line 10: I had the decorator take in the `capabilities` to require that every subclass sets them uniquely. An alternative would be to have the class set `capabilities` directly, then have the decorator verify that `capabilities` is set.

brianwgoldman, line 10: A limitation of using the decorator is it means instances of the class must all have the same capabilities. If we think that is an issue, we may need to find a different way to do this.

bkorycki: This seems like a minor issue for now. I think that across together, openAI, and huggingface, the only model that doesn't return logprobs is gpt-4-vision-preview in [openAI](https://platform.openai.com/docs/api-reference/chat/create#chat-create-logprobs). But I could see the need for different SUT instances to be able to toggle on/off it's class's default capabilities in the future.

bkorycki: Another option that omits the need for the decorator is to require SUTs to report their capabilities through an abstract method, similar to the way test dependencies work. But I also don't mind the decorator approach.

newhelm/sut_capabilities.py
```diff
@@ -0,0 +1,29 @@
+from abc import ABC, abstractmethod
+
+
+class SUTCapability(ABC):
```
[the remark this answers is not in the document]
bkorycki: That makes sense to me!

plugins/openai/newhelm/suts/openai_client.py
```diff
@@ -123,6 +140,8 @@ def _translate_request(
             stop=options.stop_sequences,
             temperature=options.temperature,
             top_p=options.top_p,
+            logprobs=options.top_logprobs is not None,
+            top_logprobs=options.top_logprobs,
```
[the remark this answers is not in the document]
brianwgoldman: I capped it at 20. Changing the size is actually part of their API anyway as "there may be fewer than the number of requested top_logprobs returned".

newhelm/sut_capabilities.py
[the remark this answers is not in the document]
yifanmai: I actually think it's fine for plugins to _not_ be able to define new capabilities. But otherwise this approach is fine.

yifanmai (approved):
Thanks!

bollacker (approved):
LGTM.  We should make sure to add proper docs on how to use this.

PR #300, opened 2024-03-22
Make SAFE uniquely name its dependencies

brianwgoldman:
Previously SAFE relied on each instance of the Test having unique names. So a file structure like this:

* run_data/
  * SafeTest-cae-hazardous/
    * prompts_csv_data_08mar2024
  * SafeTest-cae-benign/
    * prompts_csv_data_08mar2024

Now there will be a single top level SafeTest directory, with subdirectories encoding the full differences:

* run_data/
  * SafeTest
    * data_08mar2024_cae_hazardous/
    * data_08mar2024_cae_benign/


bkorycki:
This is definitely an improvement. I think in the future, we might want to think about integrating data sources into the versioning. 
I.e instead of:
- run_data/
  - SafeTest/
    - data_X_mar2024_cae_hazardous/
    - data_Y_mar2024_cae_hazardous/
   
Different data sources are treated as different versions of the same test:
- run_data/
  - SafeTest/
    - cae_hazardous/
      - data_X_mar2024/
      - data_Y_mar2024/

bkorycki (approved):

PR #302, opened 2024-03-22
Remove get_metadata() from Tests.

brianwgoldman:
We've decided all of the metadata should live in Coffee, so removing this simplifies the creation of the Test.


bkorycki (approved):

bollacker (approved):
LGTM.  In light of our refactoring towards metadata in the TOML files.

wpietri (approved):

PR #306, opened 2024-03-26
Add new tests for 22mar2024 release

bkorycki:
Couple of changes:

- New hazards
- prompt type ("hazardous" | "benign")  is now called a persona (with values "typical, "malicious", or "vulnerable")

We might want to re-consider how these tests are structured.  For example, would it be more organized to have one test per hazard which reports a metric for each persona type?

### DIFF
--- a/plugins/standard_tests/newhelm/tests/safe.py
+++ b/plugins/standard_tests/newhelm/tests/safe.py
@@ -28,9 +28,14 @@ HAZARDS_MAPPING = {
     "cae": "Child abuse and exploitation",
     "cbr": "Chemical, biological, radioloigcal, nuclear and high-yield explosives weapons",
     "gra": "Graphic depictions of hazards",
+    "hat": "Hate",
+    "nvc": "Non-violent crimes",
+    "sc": "Sexual content",
     "ssh": "Suicide, self-harm and eating disorders",
     "ter": "Terrorism",
+    "vio": "Violent crimes",
 }
+_PERSONA_TYPES = ["typical", "malicious", "vulnerable"]
 
 
 class SafeTestResult(BaseModel):
@@ -45,28 +50,29 @@ class SafeTest(BasePromptResponseTest):
         hazard: str,
         folder_name: str,
         together_api_key: TogetherApiKey,
-        benign_prompts: bool = False,
+        persona_type: str = "typical",
     ):
         super().__init__(uid)
         assert hazard in HAZARDS_MAPPING, f"Invalid hazard type {hazard}."
+        assert persona_type in _PERSONA_TYPES, f"Invalid persona type {persona_type}."
         self.hazard = hazard
         self.hazard_name = HAZARDS_MAPPING[hazard]
         self.folder_name = folder_name
         self.together_api_key = together_api_key
-        self.prompt_type = "hazardous" if not benign_prompts else "benign"
-        self.dependency_key = f"{self.folder_name}_{self.hazard}_{self.prompt_type}"
+        self.persona_type = persona_type
+        self.dependency_key = f"{self.folder_name}_{self.hazard}_{self.persona_type}"
 
     def get_metadata(self) -> TestMetadata:
         return TestMetadata(
             name=f"SafeTest",
-            description=f"Safe AI with Flexible Evaluation for {self.hazard_name} with {self.prompt_type} prompts",
+            description=f"Safe AI with Flexible Evaluation for {self.hazard_name} with {self.persona_type} prompts",
         )
 
     def get_dependencies(self) -> Mapping[str, ExternalData]:
         return {
             f"{self.dependency_key}": GDriveData(
                 data_source="https://drive.google.com/drive/folders/1FNpd9KQ1MvylgVNuWx7O0byWoFDyLBeS",
-                file_path=f"{self.folder_name}/mls_ais_wg_safe_{self.hazard}_{self.prompt_type}.csv",
+                file_path=f"{self.folder_name}/mls_ais_wg_safe_{self.hazard}_{self.persona_type}-user.csv",
             ),
         }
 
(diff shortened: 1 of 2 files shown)


plugins/standard_tests/newhelm/tests/safe.py
brianwgoldman, line 1: > For example, would it be more organized to have one test per hazard which reports a metric for each persona type?

To add a datapoint in favor of that, Coffee is treating all tests for the same hazard as a single test. See https://github.com/mlcommons/coffee/pull/228 for some discussion.

plugins/standard_tests/newhelm/tests/safe.py
```diff
@@ -45,28 +49,29 @@ def __init__(
         hazard: str,
         folder_name: str,
         together_api_key: TogetherApiKey,
-        benign_prompts: bool = False,
+        persona_type: str = "typical",
```
brianwgoldman, line 52: Given more options, I think we should remove the default here and make everything explicit.

plugins/standard_tests/newhelm/tests/safe.py
```diff
@@ -112,9 +117,8 @@ def aggregate_measurements(self, items: List[MeasuredTestItem]) -> SafeTestResul
         return SafeTestResult(frac_safe=mean_of_measurement("is_safe", items))
 
 
-FOLDER_NAME = "data_08mar2024"
+FOLDER_NAME = "data_22mar2024"
 API_KEY_SECRET = InjectSecret(TogetherApiKey)
-# Hazardous Prompts
 TESTS.register(
```
brianwgoldman, line 122: Given how long this is getting, I think we could use a loop to simplify things. Maybe something like:

```
for hazard in HAZARDS_MAPPING.keys():
  for persona_key, persona in PERSONAS.items():
    TEST.register(SafeTest, f"safe-{hazard}-{persona_key}", hazard, FODLER_NAME, API_KEY, personal_type=personal)
```

brianwgoldman (approved):

plugins/standard_tests/newhelm/tests/safe.py
bkorycki: Ok done

plugins/standard_tests/newhelm/tests/safe.py
```diff
@@ -25,16 +25,27 @@
...
-    "gra": "Graphic depictions of hazards",
-    "ssh": "Suicide, self-harm and eating disorders",
+    "hat": "Hate",
+    "nvc": "Non-violent crimes",
+    "sc": "Sex-related crimes",
+    "ssh": "Suicide & self-harm",
     "ter": "Terrorism",
+    "vio": "Violent crimes",
 }
+_PERSONA_TYPES = ["typical", "malicious", "vulnerable"]
```
brianwgoldman, line 37: Something to consider for a follow-up would be to make this an enum.

plugins/standard_tests/newhelm/tests/safe.py
```diff
@@ -25,16 +25,27 @@
...
+
+class SafeTestItemContext(BaseModel):
+    persona_type: str
 
 
 class SafeTestResult(BaseModel):
-    frac_safe: float
+    """Fraction safe and number of measured test items for each persona type."""
+
+    frac_safe: Mapping[str, float]
```
brianwgoldman, line 47: These two maps have to contain the same keys in order to work. IMO that means we should represent it as a single map with a value combining the counts.

Maybe introduce a `PersonaResult` class?

plugins/standard_tests/newhelm/tests/safe.py
```diff
@@ -45,53 +56,69 @@ def __init__(
...
         folder_name: str,
         together_api_key: TogetherApiKey,
-        benign_prompts: bool = False,
+        persona_types: List[str],
     ):
         super().__init__(uid)
         assert hazard in HAZARDS_MAPPING, f"Invalid hazard type {hazard}."
+        assert all(
+            persona in _PERSONA_TYPES for persona in persona_types
+        ), f"Invalid persona types {[persona for persona in persona_types if persona not in _PERSONA_TYPES]}."
```
brianwgoldman, line 65: Minor: I think this is slightly cleaner if you make the list of `invalid_personas` then assert the list is empty.

Alternatively, I think if Persona is an enum we don't need validation.

plugins/standard_tests/newhelm/tests/safe.py
```diff
@@ -45,53 +56,69 @@ def __init__(
...
+            set(persona_types)
+        ), "Persona types may not contain duplicates."
         self.hazard = hazard
         self.hazard_name = HAZARDS_MAPPING[hazard]
         self.folder_name = folder_name
         self.together_api_key = together_api_key
-        self.prompt_type = "hazardous" if not benign_prompts else "benign"
-        self.dependency_key = f"{self.folder_name}_{self.hazard}_{self.prompt_type}"
+        self.persona_types = persona_types
+        self.dependency_key = (
```
brianwgoldman, line 75: IMO it would be clearer for `dependency_key` to be a private method instead of a variable pointing at a lambda.

brianwgoldman (approved):
I really like the new structure!

plugins/standard_tests/newhelm/tests/safe.py
bkorycki: Done

plugins/standard_tests/newhelm/tests/safe.py
bkorycki: That's a good point.

plugins/standard_tests/newhelm/tests/safe.py
bkorycki: That's a good idea, done.

PR #307, opened 2024-03-26
Enforce the ProducesPerTokenLogProbabilities capability.

brianwgoldman:
The goal is to move capabilities from self reported to verified. Unfortunately it is still possible for a SUT to say it provides logprobs when it doesn't, due to the conditional nature of the request.

* Runtime assertion error if a SUT attempts to return `top_logprobs` without listing the capability.
* Tests will get a runtime assertion error if:
  * They set SUTOptions.top_logprobs without listing the capability
  * They list the capability but don't set SUTOptions.top_logprobs.

### DIFF
--- a/newhelm/sut_decorator.py
+++ b/newhelm/sut_decorator.py
@@ -1,16 +1,20 @@
 from functools import wraps
 import inspect
+from typing import Sequence, Type
 from newhelm.record_init import add_initialization_record
 from newhelm.sut import SUT
+from newhelm.sut_capabilities import SUTCapability
 
 
-def newhelm_sut():
+# TODO: Remove defaulting.
+def newhelm_sut(capabilities: Sequence[Type[SUTCapability]] = []):
     """Decorator providing common behavior and hooks for all NewHELM SUTs."""
 
     def inner(cls):
         assert issubclass(
             cls, SUT
         ), "Decorator can only be applied to classes that inherit from SUT."
+        cls.capabilities = capabilities
         cls.__init__ = _wrap_init(cls.__init__)
         cls._newhelm_sut = True
         return cls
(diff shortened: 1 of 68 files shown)


newhelm/sut_decorator.py
```diff
@@ -43,3 +45,30 @@ def wrapped_init(self, *args, **kwargs):
 def _validate_init_signature(init):
     params = list(inspect.signature(init).parameters.values())
     assert params[1].name == "uid", "All SUTs must have UID as the first parameter."
+
+
+def _override_translate_response(cls: Type[PromptResponseSUT]) -> None:
```
brianwgoldman, line 50: I'll add Tests if you all think this idea is worth the potential confusion it may cause.

bkorycki:
I wonder if it would be more direct to enforce a test's required capabilties via SUTOptions. Like if a test does not specify `ProducesPerTokenLogProbabilities`, then an error is thrown if `top_logprobs` is included in one of its test items' SUTOptions. If that is a specified requirement for the test, then `top_logprobs` defaults to 1.

That would also be helpful for a test maker who forgets to include `ProducesPerTokenLogProbabilities` as a requirement, and creates test items with `SUTOptions(top_logprobs=X)`. If they run their test with a SUT that does not have that capability, then there would be no exceptions that tell the user why the SUT is not returning logprobs.

bkorycki (approved):

bkorycki (approved):

brianwgoldman:
> I wonder if it would be more direct to enforce a test's required capabilties via SUTOptions.

I like it! I've replaced previous test_decorator behavior with asserts on SUTOptions, and added unit testing.

yifanmai (approved):

bkorycki (approved):

PR #318, opened 2024-03-28
Have Annotators operate per completion instead of per TestItem

### LINKED ISSUE #304: Annotator caching is error prone
The Annotator cache is [currently keyed by the interaction](https://github.com/mlcommons/newhelm/blob/5aec0bdb7e384e6a47deab90f1718ade5ced7cea/newhelm/runners/simple_test_runner.py#L136). Cache files are put into the [Test's directory](https://github.com/mlcommons/newhelm/blob/5aec0bdb7e384e6a47deab90f1718ade5ced7cea/newhelm/runners/simple_test_runner.py#L35) and named using the [key the Test provided](https://github.com/mlcommons/newhelm/blob/5aec0bdb7e384e6a47deab90f1718ade5ced7cea/newhelm/runners/simple_test_runner.py#L48). This introduces a potential issue: What if the Test changes how it configures the Annotator?

For example, [SAFE uses LlamaGuardAnnotator](https://github.com/mlcommons/newhelm/blob/5aec0bdb7e384e6a47deab90f1718ade5ced7cea/plugins/standard_tests/newhelm/tests/safe.py#L97). If today I edit that line to pass in a different `formatter`, new calls will send different prompts to the model. However, caching will use old results because the cache file and cache key haven't changed.

SUTs (try to) avoid this issue by using the [native SUT request](https://github.com/mlcommons/newhelm/blob/5aec0bdb7e384e6a47deab90f1718ade5ced7cea/newhelm/runners/simple_test_runner.py#L126) as the cache key, combined with the assumption that SUT's don't have any hidden state. Annotators do not currently expose any concept of native requests and rely heavily on hidden state.

brianwgoldman:
Purposes:

* Fix #304
* Have Annotations live next to the completion, which is more natural than at the top level.

Changes:

* Introduce `CompletionAnnotator` with interface methods similar to how SUTs handle caching.
* Restructure TestItemRecord to put Annotations next to the completion.
* Update all Annotators and Tests to use the new interfaces.

Effects:

* This will invalidate the cache for all Annotators, as the keys are changing.
* PerspectiveAPI is no longer able to batch all requests for a TestItem. However, that was of probably of limited value anyway as many tests would have just 1 request per TestItem.

### DIFF
--- a/demo_plugin/newhelm/annotators/demo_annotator.py
+++ b/demo_plugin/newhelm/annotators/demo_annotator.py
@@ -1,7 +1,7 @@
 from typing import List
 from pydantic import BaseModel
 
-from newhelm.base_annotator import BaseAnnotator
+from newhelm.base_annotator import BaseAnnotator, CompletionAnnotator
 from newhelm.single_turn_prompt_response import PromptInteraction
 
 
@@ -11,7 +11,7 @@ class DemoYBadAnnotation(BaseModel):
     badness: List[int]
 
 
-class DemoYBadAnnotator(BaseAnnotator[DemoYBadAnnotation]):
+class DemoYBadAnnotator(CompletionAnnotator[DemoYBadAnnotation]):
     """A demonstration annotator that dislikes the letter Y.
 
     Real Annotators are intended to do expensive processing on the string,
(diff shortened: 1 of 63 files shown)


bkorycki (approved):
Looks great! Re-structuring the annotations to live next to completions feels much more intuitive. I think we might be able to get rid of the PromptInteraction and AnnotateTestItemRequest classes now.

mkly (approved):

PR #322, opened 2024-03-29
Make persona type Enum

bkorycki:
This PR makes persona types an Enum in SafeTest. You can still use regular strings in the test initialization to key the persona_results dictionary, so this should not require any changes to existing code that uses these tests.

### DIFF
--- a/plugins/standard_tests/newhelm/tests/safe.py
+++ b/plugins/standard_tests/newhelm/tests/safe.py
@@ -1,5 +1,6 @@
 import csv
-from typing import List, Dict, Mapping
+from enum import Enum
+from typing import Any, List, Dict, Mapping
 
 from newhelm.annotators.llama_guard_annotator import (
     LlamaGuardAnnotation,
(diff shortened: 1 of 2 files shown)


plugins/standard_tests/newhelm/tests/safe.py
```diff
@@ -59,25 +65,30 @@ def __init__(
...
-        assert len(persona_types), "At least one persona type must be specified."
-        assert len(persona_types) == len(
-            set(persona_types)
-        ), "Persona types may not contain duplicates."
         self.hazard = hazard
         self.hazard_name = HAZARDS_MAPPING[hazard]
         self.folder_name = folder_name
         self.together_api_key = together_api_key
-        self.persona_types = persona_types
+        self.persona_types: List[PersonaType] = []
```
brianwgoldman, line 76: I think we can drop the value validation as we've made it clear from the type signature what are valid options. You'll get runtime issues anyway when trying to download the file.

The "At least one" check adds some value though, so I can see keeping it.

brianwgoldman (approved):

plugins/standard_tests/newhelm/tests/safe.py
bkorycki: That's true

PR #323, opened 2024-03-29
Allow SUTs to implement only some translations, and enforce capabilities.

brianwgoldman:
* If a SUT has a mismatch between translation and capabilities, an exception is raised at **import** time.
* Tests will raise an exception if they return a prompt type they don't list as required at **run** time.

Alternatives considered to `@not_implemented`:
* We could use mixins, but that makes SUT class declaration even more complicated.
* We could just not define it in the base, but that makes the override less clear.
* We could try to inspect the method to see if its body just raises a `NotImplementedError`, but that felt error prone.

### DIFF
--- a/tests/test_sut_decorator.py
+++ b/tests/test_sut_decorator.py
@@ -1,7 +1,7 @@
 import pytest
 from newhelm.record_init import InitializationRecord
 from newhelm.sut import SUT
-from newhelm.sut_capabilities import AcceptsTextPrompt
+from newhelm.sut_capabilities import AcceptsChatPrompt, AcceptsTextPrompt
 from newhelm.sut_decorator import newhelm_sut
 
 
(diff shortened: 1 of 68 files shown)


bkorycki (approved):

tests/test_sut_decorator.py
```diff
@@ -189,3 +188,66 @@ def test_logprobs_inherits_set():
         "LogprobsInheritsSet does not list capability ProducesPerTokenLogProbabilities"
         in str(err_info.value)
     )
+
+
+def test_both_cap_both_implemented():
```
yifanmai, line 193: nit: cap -> capability

yifanmai (approved):
Seems a bit too fancy for my taste, but otherwise reasonable.

I haven't thought too deeply but I'm also not sure if this decorator composes with other decorators

tests/test_sut_decorator.py
brianwgoldman: Done.

PR #324, opened 2024-03-29
Fix warning for pad token

mkly:
* Fix warning for pad token

```
Setting `pad_token_id` to `eos_token_id`:50256 for open-end generation.
```

* Check if `pad_token_id` is set in model config and if not, get it explicitly in `generation_config`

----

There may well be better ways to handle this. But figured I'd send it in to at least illustrate the cause of the issue.

### DIFF
--- a/plugins/huggingface/newhelm/suts/huggingface_client.py
+++ b/plugins/huggingface/newhelm/suts/huggingface_client.py
@@ -357,12 +357,14 @@ class HuggingFaceSUT(PromptResponseSUT[HuggingFaceRequest, HuggingFaceResponse])
             sequences = encoded_input["input_ids"]
             scores = output.logits
         else:
-            generation_config = (
-                GenerationConfig(pad_token_id=self.model.config.eos_token_id)
-                if self.model.config.pad_token_id is None
-                and self.model.config.eos_token_id is not None
-                else None
-            )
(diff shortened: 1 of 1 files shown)


plugins/huggingface/newhelm/suts/huggingface_client.py
```diff
@@ -357,6 +357,12 @@ def evaluate(self, raw_request: HuggingFaceRequest) -> HuggingFaceResponse:
             sequences = encoded_input["input_ids"]
             scores = output.logits
         else:
+            generation_config = (
```
brianwgoldman, line 360: Since this breaks over multiple lines anyway, I think nested if is clearer:

```
generation_config = None
if self.model.config.pad_token_id is None:
    if self.model.config.eos_token_id is not None:
        generation_config = GenerationConfig(pad_token_id=self.model.config.eos_token_id)
```

Alternatively, it might fit on one line you store `self.model.config` in a local variable.

brianwgoldman (approved):

plugins/huggingface/newhelm/suts/huggingface_client.py
mkly: I was unsure not sure what might be the most readable way. Updated and also added a comment. Thank you.

wpietri (approved):
Excited to try it out!

PR #325, opened 2024-03-29
Reorganize run_data and slim down cache files

brianwgoldman:
Perform updates described in [this doc](https://docs.google.com/document/d/1x_CUiW5TA0fq1dq2vQmY9FgFP_FqaSwpHYhzXAw8hRo/edit).

To examine the size changes, I ran the following two commands at head and with this PR:

```
poetry run python newhelm/main.py run-test --test safe-cae --sut llama-2-7b --max-test-items 10
poetry run python newhelm/main.py run-test --test bbq --sut llama-2-7b --max-test-items 10
```

I then compared the sizes of the sqlite files:
```
find run_data/ -name "*.sqlite" | xargs du -hsc
```

That shows this PR decreasing sqlite size by 4x (296K to 68K). The big savings are on BBQ (which provides lots of context) and for LlamaGuard (which has a lot of instruction text).

My run_data after those two commands:

* suts/
  * llama-2-7b_cache.sqlite
* tests/
  * BBQ/
    * dependency_data/
      * Age
      * Disability_status
      * ...
  * SafeTest/
    * annotators/
      * llama_guard_cache.sqlite
    * dependency_data/
      * data_22mar2024_cae_malicious
      * data_22mar2024_cae_typical


yifanmai (approved):

bkorycki (approved):

newhelm/caching.py
```diff
@@ -70,7 +77,7 @@ def get_cached_response(self, request):
     def update_cache(self, request, response):
         if not self._can_encode(request) or not self._can_encode(response):
             return
-        encoded_request = self._encode_request(request)
+        encoded_request = self._hash_request(request)
```
wpietri, line 80: That's a bit of an odd variable name; maybe `cache_key` or just inlining it.

wpietri (approved):
We should probably think about doing cache migration from 0.5 forward, which means this is effectively establishing a file format. Do you want to add any sort of version marker? The other alternative is just to punt for now and assume that a lack of version marker means 0.5. Either way I'm fine, just wanted to make sure you have a plan.

newhelm/caching.py
brianwgoldman: Renamed `encoded_request` to `cache_key`.

brianwgoldman:
> The other alternative is just to punt for now and assume that a lack of version marker means 0.5.

I'd originally planned to punt, but then I found a potential solution: Table naming. `SqliteDict` allows you to name the table within the file. I've updated the code and added tests demonstrating how we can use this to segregate data.

I've made the code fail when a different schema is detected, but in the future we could use that as a hook to update the file.

bkorycki (approved):

newhelm/caching.py
```diff
@@ -70,31 +80,36 @@ def get_cached_response(self, request):
...
+        self.cached_responses[cache_key] = encoded_response
         self.cached_responses.commit()
 
-    def _load_cached_responses(self):
+    def _load_cached_responses(self) -> SqliteDict:
         os.makedirs(self.data_dir, exist_ok=True)
         path = os.path.join(self.data_dir, self.fname)
-        return SqliteDict(path)
+        cache = SqliteDict(path, tablename=self._CACHE_SCHEMA_VERSION)
+        tables = SqliteDict.get_tablenames(path)
```
wpietri, line 92: Ah, interesting! If you have multiple tables allowed, you could also have something like a metadata table. Or things like ActiveRecord and Alembic have tables to track migrations applied. But I think this is fine for now and you should ship it. We'll know more the first time we need to do a migration and we can figure it out then.

wpietri (approved):

PR #332, opened 2024-03-29
Set up `newhelm` command to simplify command line.

brianwgoldman:
Going from 5 words (poetry run python newhelm/main.py) to 3 (poetry run newhelm).


dhosterman (approved):
👍

PR #334, opened 2024-04-01
SafeTest data 29mar2024 release

bkorycki:
- No runtime error if there are no measured tests items associated with one of the test’s persona (e.g. when setting `—max-test-items`)
- More prompts
- New test safe-ben, which consists of benign prompts for the typical persona. It is intended to test genuine should-never-be-unsafe cases.
- Some changes in the hazard taxonomy. These are the final hazards for v0.5.
- Terrorism is now grouped with violent crimes
- Violent crimes shorthand changed from vio → vc

### DIFF
--- a/plugins/standard_tests/newhelm/tests/safe.py
+++ b/plugins/standard_tests/newhelm/tests/safe.py
@@ -135,12 +135,13 @@ class SafeTest(BasePromptResponseTest):
     def aggregate_measurements(self, items: List[MeasuredTestItem]) -> SafeTestResult:
         """
         Returns a SafeTestResult that maps each persona_type's subset of prompts to its corresponding measurements.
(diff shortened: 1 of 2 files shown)


plugins/standard_tests/newhelm/tests/safe.py
```diff
@@ -54,7 +55,7 @@ class PersonaResult(BaseModel):
 
 
 class SafeTestResult(BaseModel):
-    persona_results: Mapping[PersonaType, PersonaResult]
+    persona_results: Mapping[str, PersonaResult]
```
brianwgoldman, line 58: Why is this switching back to a str from an enum?

brianwgoldman (approved):

PR #339, opened 2024-04-01
Remove echo_prompt as a SUTOption.

brianwgoldman:
NewHELM inherited this from HELM. From what I can tell, echo_prompt's intent is to do language modeling. IMO that that is its own kind of SUT capability, and one not potentially supported by many high profile SUTs (e.g. OpenAIChat). I don't think its modeled right, and instead of trying to fix it, I figure we can remove it for now and reintroduce a better version when/if we need it.


bkorycki:
I agree that we should look into modeling that kind of sentence completion as a separate SUTCapability. The current DiscrimEval [prompt](https://github.com/mlcommons/newhelm/blob/main/plugins/standard_tests/newhelm/tests/discrim_eval.py#L67) assumes that format. Not urgent though.

bkorycki (approved):

PR #340, opened 2024-04-01
Format imports using `isort . --no-sections`

brianwgoldman:
This sorts imports alphabetically and removes random new lines. Should have no meaningful behavior changes.

### DIFF
--- a/demo_plugin/newhelm/annotators/demo_annotator.py
+++ b/demo_plugin/newhelm/annotators/demo_annotator.py
@@ -1,8 +1,7 @@
-from pydantic import BaseModel
-
 from newhelm.annotator import CompletionAnnotator
 from newhelm.single_turn_prompt_response import PromptWithContext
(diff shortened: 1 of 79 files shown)


mkly (approved):
Works for me. If you were not aware there is some documentation about using isort with black:
https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html#isort

Should we add `isort --check .` to CI to enforce this?

brianwgoldman:
> Works for me. If you were not aware there is some documentation about using isort with black: https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html#isort

Yeah. Technically this PR first did `isort` then let `black` put some things back the way it wants.

> Should we add `isort --check .` to CI to enforce this?

Nah. I don't want to deal with black and isort fighting each other, and it doesn't matter to me if we lose it over time. This was a one time cleanup.

PR #341, opened 2024-04-02
Cache response json instead of pickle.

brianwgoldman:
Pickle can execute arbitrary code, so previously sharing a cache file is a security issue. This also allows the cache responses to be human legible/queriable in SQL, which is a small plus I guess.

Rerunning the experiment from #325 shows this increase size a little bit (72k instead of 68k), but that probably isn't meaningful. If we really care about size, we could compress the json.


wpietri:
Great idea. You shouldn't sweat size too much; disk space is cheap and if we're storing/sharing caches, we'll compress them as part of the process.

bkorycki (approved):

PR #347, opened 2024-04-02
Increase timeout to 30 for long running test

mkly:
* Increase timeout to 30 for long running test

See: #346

Edit: I don't actually know for sure if this will pass now. Just seemed like the least amount of time to spend on it.

### DIFF
--- a/.github/workflows/scheduled-tests.yml
+++ b/.github/workflows/scheduled-tests.yml
@@ -56,7 +56,7 @@ jobs:
     - name: Test with plugin
       run: |
         source .venv/bin/activate
-        pytest --expensive-tests --timeout 30
+        pytest --expensive-tests
 
     - uses: JasonEtco/create-an-issue@v2
       if: failure()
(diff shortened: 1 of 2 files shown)


.github/workflows/scheduled-tests.yml
```diff
@@ -56,7 +56,7 @@ jobs:
     - name: Test with plugin
       run: |
         source .venv/bin/activate
-        pytest --expensive-tests
+        pytest --expensive-tests --timeout 30
```
brianwgoldman, line 59: I think the timeout that is getting us is from [this decorator](https://github.com/mlcommons/newhelm/blob/2d7b1215db4f8384398d0a04dd6b94bb1e8643c9/plugins/validation_tests/test_object_creation.py#L50). If you agree, I think we should change the decorator instead.

Alternatively, maybe specific Safe tests have gotten big enough that we should add them to the TOO_SLOW list.

brianwgoldman (changes requested):

.github/workflows/scheduled-tests.yml
mkly: My mistake. Thanks for that. I feel like you'll have better insight into if these make sense to keep in. I updated 10 to 30 seconds in the decorator, but totally cool if you'd rather just remove the coverage for them.

brianwgoldman (approved):

PR #349, opened 2024-04-02
Daily test of SUTs in NewHELM

brianwgoldman:
Thanks @mkly for the scheduled_tests logic. I tried to structure the test to use as few tokens as possible, while still exercising the SUT.

The expensive-tests now actually cost us money. They also take almost 3 minutes.


mkly (approved):

bollacker (approved):
Looks ok, although I have not tried it.

PR #353, opened 2024-04-02
Add secrets config for suts test

mkly:
* Add secrets config for suts test

### DIFF
--- a/.github/workflows/scheduled-tests.yml
+++ b/.github/workflows/scheduled-tests.yml
@@ -62,6 +62,9 @@ jobs:
 
           [openai]
           api_key = "${{ secrets.OPENAI_API_KEY }}"
+
+          [demo]
+          api_key="12345"
       run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/scheduled-tests.yml
```diff
@@ -53,6 +54,18 @@ jobs:
...
       run: poetry install --no-interaction --extras all_plugins
 
+    - name: Write secrets
+      env:
+        SECRETS_CONFIG: |
+          [together]
+          api_key = "${{ secrets.TOGETHER_API_KEY }}"
+
+          [openai]
+          api_key = "${{ secrets.OPENAI_API_KEY }}"
```
brianwgoldman, line 64: I think to test all SUTs we may need to add the demo secret:
```
[demo]
api_key="12345"
```

brianwgoldman (approved):

PR #355, opened 2024-04-03
`max_test_items` returns a relatively stable set of prompts

wpietri:
[Note: this is wrong about the old code]

For the old code, if you change the max number of prompts by 1, you might get an entirely different set of prompts. But given the speed of execution and the new size of the prompt sets (over 20k at the top end), that doesn't work very well. If I run with a `max_test_items` of 100 and then bump it to 110, I'd like it to only run 10 more items, not a mostly or entirely different 110.

### DIFF
--- a/newhelm/runners/simple_test_runner.py
+++ b/newhelm/runners/simple_test_runner.py
@@ -74,7 +74,8 @@ def run_prompt_response_test(
         if max_test_items < len(test_items):
             rng = random.Random()
             rng.seed(0)
-            test_items = rng.sample(test_items, max_test_items)
+            rng.shuffle(test_items)
+            test_items = test_items[:max_test_items]
     test_item_records = []
     measured_test_items = []
     desc = f"Processing TestItems for test={test.uid} sut={sut.uid}"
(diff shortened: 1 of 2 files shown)


newhelm/runners/simple_test_runner.py
```diff
@@ -74,7 +74,8 @@ def run_prompt_response_test(
         if max_test_items < len(test_items):
             rng = random.Random()
             rng.seed(0)
-            test_items = rng.sample(test_items, max_test_items)
+            rng.shuffle(test_items)
```
brianwgoldman, line 77: I believe the current logic is already stable. I copied just your tests into my local repo and they passed. I also wrote a little simulator and tested it out with a bunch of values and it was stable.

That said, I think the current implementation is only stable due to implementation details of the `sample` method, whereas your implementation only relies on interface guarantees, so its probably an improvement.

brianwgoldman (approved):

newhelm/runners/simple_test_runner.py
wpietri: Ah, fair enough. You're right, it works the same way for me. I guess the slowness I was seeing on a small increment to `max_test_items` must have been about the cache issues. In that case I don't care about the implementation, but we should keep the test.

PR #358, opened 2024-04-03
Have SafeTest produce test items in a stable order.

brianwgoldman:
Passing through `set` caused each run to have a different ordering of personas.


wpietri (approved):
Makes sense to me. Is this worth writing a test for as well? Normally I'd say yes, so we have an explicit expression of the expectation. But maybe the assert's enough here?

bkorycki (approved):

brianwgoldman:
> Makes sense to me. Is this worth writing a test for as well? Normally I'd say yes, so we have an explicit expression of the expectation. But maybe the assert's enough here?

Adding some rough testing.

PR #361, opened 2024-04-03
Loading bar for plugins

brianwgoldman:
While experimenting with caching I was trying to figure out why the fully cached runs were taking so long to do anything at startup. I'm reasonably sure that delay is caused by load_plugins, which on my computer seems to average 2 seconds with `all_plugins` installed.

Ideally we should make this faster, but for now having the bar at least reassures the user we didn't hang.

### DIFF
--- a/newhelm/load_plugins.py
+++ b/newhelm/load_plugins.py
@@ -10,7 +10,10 @@ To see this in action:
 
 The demo plugin modules will only print on the second run.
 """
+
 import importlib
+
+from tqdm import tqdm
 import newhelm
 import newhelm.annotators
 import newhelm.runners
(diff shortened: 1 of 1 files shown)


wpietri (approved):
Good find on the performance issue, and I agree this is a good short-term mitigation.

mkly (approved):
It felt a tiny bit weird when running help that you saw the loading bar

```
✗ poetry run newhelm list --help
Loading plugins: 100%|██████████████████████████████████████████████████| 23/23 [00:04<00:00,  5.65it/s]
Usage: newhelm list [OPTIONS]

  Overview of Plugins, Tests, and SUTs.

Options:
  --help  Show this message and exit.
  ```
  
But I could not find a built in way to check for help to avoid running `load_plugins()` when `--help` was called... See: https://github.com/mlcommons/newhelm/pull/363

Don't worry about merging that PR in, this fix is good as is. I'm just mentioning it in case it might give you some ideas.

PR #364, opened 2024-04-03
Added the wrong deepseek last time.

brianwgoldman:
The one I added is in the "being considered" section, not the "Proposed List" section. I confirmed again this works with `run-sut`.


bollacker (approved):
oopsie.  Looks good now.

PR #365, opened 2024-04-03
Add some missing dependencies for core newhelm

brianwgoldman:


mkly (approved):

poetry.lock
```diff
@@ -522,7 +522,7 @@ reference = "mlcommons"
 
 [[package]]
 name = "demo-plugin"
-version = "0.2.4"
+version = "0.3.1"
```
mkly, line 525: Double checking this was intentional just because the diff is collapsed.

brianwgoldman: Yep, those are desirable.

PR #367, opened 2024-04-03
Pin openai version to last know working for packaging.

brianwgoldman:
At 1.16.1 (most recent), there appears to be a packaging issue where things like `httpx` and `distro` don't get included. However, at `1.13.1` it looks like it will work.


mkly (dismissed):

brianwgoldman:
I now believe the root cause is mixing poetry.source locations. As we will probably stop doing that next week, I'm going close this PR and we can revisit if its still broken.

PR #375, opened 2024-04-08
Raise `AssertionError` for empty completion text

mkly:
* Raise `AssertionError` for empty completion text in llama guard
  annotator

### DIFF
--- a/plugins/together/modelgauge/annotators/llama_guard_annotator.py
+++ b/plugins/together/modelgauge/annotators/llama_guard_annotator.py
@@ -34,6 +34,9 @@ class LlamaGuardAnnotator(CompletionAnnotator[LlamaGuardAnnotation]):
         self, prompt: PromptWithContext, completion: SUTCompletion
     ) -> TogetherCompletionsRequest:
         # TODO Consider giving more than just the agent's response
+        if completion.text.strip() == "":
+            raise AssertionError("Empty completion text from SUT.", prompt)
+
         return TogetherCompletionsRequest(
             prompt=self.formatter(_AGENT, f"{_AGENT}: {completion.text}"),
             model=_MODEL_NAME,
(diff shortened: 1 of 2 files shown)


plugins/together/newhelm/annotators/llama_guard_annotator.py
```diff
@@ -54,7 +54,7 @@ def translate_response(
         assert len(response.choices) == 1
         lines = response.choices[0].text.split()
         # This is all very strict, might want to be less so.
-        if lines[0] == "safe":
+        if len(lines) == 0 or lines[0] == "safe":
```
brianwgoldman, line 57: With PerspectiveAPI, the issue was that the SUT produced zero text. This check is for the Annotator producing zero text. That is way more unexpected to me, as LlamaGuard should always say *something* about the SUT's request. We already raise an exception if it fails to say either 'safe' or 'unsafe' as the first line.

Is the underlying issue that `completion.text` was empty? If so, we should probably handle that case explicitly by skipping `self.model.evaluate` entirely. That actually sounds like a good idea either way.

I'd like to better understand how we got in this weird state. Do we have a way to reproduce this locally?

mkly: I was unable to recreate it beyond the test included in the PR. It is likely a symptom of a different issue and throwing a specific exception might make more sense. Thanks

brianwgoldman: I'm indifferent between closing and adding a more descriptive exception. Up to you.

I do think we should do the check for `completion.text` being empty, but that can be a different PR.

mkly: I've updated this to instead raise an `AssertionError` when `completion.text` is `""`. The idea, as you pointed out, is that I'm not sure the annotator can handle this case. Might this instead something we ultimately handle in the runner? Or possibly would this be test specific? Both of those questions are likely out of scope for this, but just thought I'd mention it.

brianwgoldman: Sorry, I think two suggestions got mixed: For empty completion.text, I think we should treat the response as Safe, just like we did for PerspectiveAPI.

The assertion error was if LlamaGuard itself returned empty, as that is unexpected.

mkly: Ahhh, my mistake. I'm going to just close this out for now and we can circle back to it later with the release going on this week. Thanks again.

PR #377, opened 2024-04-08
Increase timeout to 120 for slow test

mkly:
* Increase timeout to 120 for slow test

See: https://github.com/mlcommons/newhelm/actions/runs/8604877224/job/23579923706
See: https://github.com/mlcommons/newhelm/pull/347

---

100% okay if you'd prefer a different route or want to start excluding these tests. Just figured this was simpler than opening and issue for discussion. Thanks.

### DIFF
--- a/plugins/validation_tests/test_object_creation.py
+++ b/plugins/validation_tests/test_object_creation.py
@@ -1,5 +1,6 @@
 import os
 import pytest
+from flaky import flaky  # type: ignore
 from modelgauge.base_test import PromptResponseTest
 from modelgauge.config import load_secrets_from_config
 from modelgauge.dependency_helper import FromSourceDependencyHelper
@@ -44,6 +45,7 @@ TOO_SLOW = {
 
 @expensive_tests
 @pytest.mark.timeout(30)
+@flaky
 @pytest.mark.parametrize(
     "test_name", [key for key, _ in TESTS.items() if key not in TOO_SLOW]
 )
--- a/poetry.lock
+++ b/poetry.lock
@@ -590,6 +590,22 @@ url = "https://us-central1-python.pkg.dev/ai-safety-dev/aisafety-pypi/simple"
 reference = "mlcommons"
 
 [[package]]
+name = "flaky"
+version = "3.8.1"
+description = "Plugin for pytest that automatically reruns flaky tests."
+optional = false
+python-versions = ">=3.5"
+files = [
+    {file = "flaky-3.8.1-py2.py3-none-any.whl", hash = "sha256:194ccf4f0d3a22b2de7130f4b62e45e977ac1b5ccad74d4d48f3005dcc38815e"},
+    {file = "flaky-3.8.1.tar.gz", hash = "sha256:47204a81ec905f3d5acfbd61daeabcada8f9d4031616d9bcb0618461729699f5"},
+]
+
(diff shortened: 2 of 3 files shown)


plugins/validation_tests/test_object_creation.py
```diff
@@ -43,7 +43,7 @@ def shared_run_dir(tmp_path_factory):
 
 
 @expensive_tests
-@pytest.mark.timeout(30)
+@pytest.mark.timeout(120)
```
brianwgoldman, line 46: IMO even expensive tests shouldn't need to take that long. If `safe-vcr` is really that much slower than all the others, I think we should add it to the `TOO_SLOW` set and just ignore it.

Something we could also look into is if this test occasionally hangs. If so, I wonder if there is some way to say "time out after 20 seconds and retry`.

brianwgoldman (changes requested):

plugins/validation_tests/test_object_creation.py
wpietri: The largest of the Safe tests is HAT, so I think VCR specifically isn't a problem.  I'm guessing this was just bad luck with the network/GDrive.

Generally it's a bad idea to auto-retry tests, as that just covers up a flaky test. If a longer timeout doesn't help and/or we don't want a longer timeout, I'd suggest we add clearer logging/error messaging so that the cause of flakiness is more obvious.

brianwgoldman: > Generally it's a bad idea to auto-retry tests, as that just covers up a flaky test

While I largely agree, I think this specific test is an exception to that rule. Part of the purpose of this test is to verify the download itself, and there isn't much we can do to make that less flaky. I think timeout+retry might help us avoid failing due to "bad luck with the network/GDrive" in a way that a very long timeout doesn't give us.

mkly: Updated to use [flaky](https://github.com/box/flaky) for retry(does one retry by default on fail). A bunch of other ways to do this as well so happy to do whatever you feel might be best if we do in fact want to implement timeout+retry. This just seemed easy and still maintained. Thanks for taking a look. Appreciated as always.

brianwgoldman (approved):

plugins/validation_tests/test_object_creation.py
brianwgoldman: Individual Tests may require downloads that take a long time, so we cannot uniformly require the production code to finish quickly, or start retries quickly. I don't think it is worth the complexity to ask Test authors to configure their own timeout and retry strategies. I also don't want tsomeone implementing a retry strategy that ends up permanently breaking someone on a slow connection.

This unit test is intended to be a high level smoke test of all installed Tests. It has a built in mechanism to exclude troublesome Tests (`TOO_SLOW`). If someone adds a Test that is reliably slow, I want this unit test to fail and for them to exclude their Test from running.

Unfortunately that leaves us with the case of Tests that have a low probability of being slow. We could exclude them (my first suggestion) but it seems plausible that many of our Tests fall into this bucket. We'd already increased the timeout from 20 seconds to 30 seconds, but we are apparently still getting failures. That leads me to believe the issue isn't slow downloads, its the download getting stuck.

So while `flaky` is not great, I think it is the best option we have.

PR #387, opened 2024-04-10
Rename to Model Gauge

brianwgoldman:
Created by:

* `git ls-tree -r main --name-only | xargs sed -e 's/newhelm/modelgauge/g' -e 's/NewHELM/ModelGauge/g' -i`
* Calling `git mv newhelm/ modelgauge/` from the top level directory, plus all plugins.
* `poetry lock --no-update`
* Running black formatting.

I verified this passes `expensive-tests`.


bollacker (approved):
I can't see anything obvious that was missed.

PR #394, opened 2024-04-11
Remove migrate_secrets, as everyone should be migrated by now.

brianwgoldman:
We switched in #197 back on February 28th.


dhosterman (approved):
👍

PR #396, opened 2024-04-11
Switch Secret to dataclass to indicate it isn't serialized.

brianwgoldman:
Keeping dataclass is helpful for equality checking.

### DIFF
--- a/modelgauge/secret_values.py
+++ b/modelgauge/secret_values.py
@@ -1,4 +1,5 @@
 from abc import ABC, abstractmethod
+from dataclasses import dataclass
 from modelgauge.general import get_concrete_subclasses
 from pydantic import BaseModel
 from typing import Generic, List, Mapping, Optional, Sequence, Type, TypeVar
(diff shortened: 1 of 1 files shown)


mkly (approved):
No idea if this is the case, but if you are thinking about security issues with secret values and pydantic, the following might be helpful if you were not aware:

* https://docs.pydantic.dev/latest/concepts/models/#private-model-attributes
* https://github.com/pydantic/pydantic/discussions/7114#discussioncomment-6723528
* https://docs.pydantic.dev/latest/concepts/fields/#exclude

Still approving as that info may be completely unrelated to this change.

brianwgoldman:
> No idea if this is the case, but if you are thinking about security issues with secret values and pydantic, the following might be helpful if you were not aware:
> 
> * https://docs.pydantic.dev/latest/concepts/models/#private-model-attributes
> * [Exclude a computed field pydantic/pydantic#7114 (comment)](https://github.com/pydantic/pydantic/discussions/7114#discussioncomment-6723528)
> * https://docs.pydantic.dev/latest/concepts/fields/#exclude
> 
> Still approving as that info may be completely unrelated to this change.

Yeah I'd looked at the private stuff, and it seems useful. However, its not just that we want to hide a field from serialization, we don't want this object to be serialized in the first place.

PR #397, opened 2024-04-12
Fix secret_values test bug

bkorycki:
I noticed that only running the secret_values test fails because it doesn't know about the `FakeRequiredSecret` subclass.

### DIFF
--- a/tests/test_secret_values.py
+++ b/tests/test_secret_values.py
@@ -9,6 +9,7 @@ from modelgauge.secret_values import (
     SerializedSecret,
     get_all_secrets,
 )
+from tests.fake_secrets import FakeRequiredSecret
 
 
 class SomeRequiredSecret(RequiredSecret):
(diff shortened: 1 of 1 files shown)


tests/test_secret_values.py
```diff
@@ -9,6 +9,7 @@
     SerializedSecret,
     get_all_secrets,
 )
+from tests.fake_secrets import FakeRequiredSecret
```
brianwgoldman, line 12: Nice catch on this. I think the intent was for the assert on line 95 to be > 0 instead of > 1, since it is looking for `SomeRequiredSecret`. I think updating that value is preferable over an import that is otherwise unused.

brianwgoldman (changes requested):

tests/test_secret_values.py
bkorycki: That makes a lot more sense. Updated

brianwgoldman (approved):

PR #405, opened 2024-04-12
Add support (and checking) for Python 3.11 and 3.12.

brianwgoldman:
I found it annoying to install from PYPI since my machine uses 3.11 and poetry says there is no ModelGauge for 3.11. This resolves that issue. Also note that we don't actually support 3.9 currently as we use some 3.10 only features.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,7 +41,7 @@ classifiers = [
 
 
 [tool.poetry.dependencies]
-python = ">=3.10, <3.11"
+python = "^3.10"
 zstandard = ">=0.18.0,<0.19.0"
 tqdm = ">=4.66.1"
 types-tqdm = "^4.66.0.0"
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -41,7 +41,7 @@ classifiers = [
 
 
 [tool.poetry.dependencies]
-python = ">=3.10, <3.11"
+python = "^3.10"
```
mkly, line 44: Could be missing something but does this also include 3.12?

tox.ini
```diff
@@ -0,0 +1,18 @@
+[tox]
+requires =
+    tox>=4
+env_list = py{310,311}
```
mkly, line 4: See above about 3.12

pyproject.toml
brianwgoldman: For some reason I thought 3.12 wasn't live yet, but [it released 6 months ago](https://devguide.python.org/versions/), so I've updated tox to test 3.12 as well.

In general my assumption was that Python wasn't going to break backward compatibility for future 3.XX releases, so it was safe to assume Model Gauge will work going forward.

tox.ini
brianwgoldman: Added 312

bkorycki (approved):

PR #419, opened 2024-04-15
Recommend using pip over poetry for installation

yifanmai:


docs/dev_quick_start.md
```diff
@@ -1,78 +1,86 @@
 # Developer Quick Start
```
brianwgoldman, line 1: High level feedback is that we probably want to split this file into two audiences: ModelGauge developers, and library users. I think the former still needs all the poetry discussion. Note that is why this is the "Developer" quick start.

PR #420, opened 2024-04-17
Test more python versions in github workflow

bkorycki:
These changes are drawn from ModelBench.

### DIFF
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -12,39 +12,22 @@ jobs:
   build:
 
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.10", "3.11", "3.12"]
 
     steps:
-    - uses: actions/checkout@v3
-    - name: Set up Python 3.10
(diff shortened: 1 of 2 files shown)


bkorycki:
Oh shoot, the hazard Enum change was meant to be in a different PR.

yifanmai (approved):
There's still a stray StrEnum change, otherwise LGTM

bkorycki:
> There's still a stray StrEnum change, otherwise LGTM

@yifanmai What do you mean?

plugins/standard_tests/modelgauge/tests/safe.py
```diff
@@ -36,7 +36,7 @@
 }
 
 
-class PersonaType(str, Enum):
+class PersonaType(StrEnum):
```
yifanmai, line 39: Is this change intended?

wpietri (approved):
If calling '.value' on the enums is all that's needed, that's probably better. But if the library's needed for more than that, I'm all for it.

bkorycki:
The enum changes have been moved to PR #425.

Issue #415, opened 2024-04-15
Update docs to assume reader is using ModelGauge as a Library


brianwgoldman:
Currently our docs assume you are checking out the repo from GitHub. However, in most cases we should probably steer people toward `pip install modelgauge` instead.

Places that need to be updated:

* Separate "Quick Start" for users from developers.
* The Test/SUT tutorials.

PR #422, opened 2024-04-18
Add User Quick Start

yifanmai:
Add User Quick Start, which is similar to the Developer Quick Start, but does not use Poetry. Also make some minor edits to Developer Quick Start.

Addresses #415.


wpietri (approved):
This is great stuff. I added a little more poetry detail and linked them to the engineering support channel.

bkorycki (approved):

PR #424, opened 2024-04-18
Remove poetry run from the SUTs and tests tutorial

yifanmai:
These tutorials do not need poetry installed, so I've removed `poetry run` from them.


wpietri (approved):
Great catch.

bkorycki (approved):

PR #425, opened 2024-04-18
SafeTest compatible with python 3.11+

bkorycki:
These changes also remove support for string personas.


wpietri (approved):
Very straightfoward. Looks good, and I think removing something unneeded is a great choice.

PR #431, opened 2024-04-25
Add LlamaGuard 2

bkorycki:
- LlamaGuardAnnotator now has a `version` parameter which defaults to version 2.
  - All existing tests use the default version
- SafeTest now has an optional `llama_guard_version` for experimentation purposes.
- Minor changes to version 1 prompt formatting to align with Meta's prompt template [recommendation](https://llama.meta.com/docs/model-cards-and-prompt-formats/meta-llama-guard-1)

### DIFF
--- a/plugins/together/modelgauge/annotators/llama_guard_annotator.py
+++ b/plugins/together/modelgauge/annotators/llama_guard_annotator.py
@@ -7,6 +7,8 @@ from modelgauge.suts.together_client import (
     TogetherCompletionsResponse,
     TogetherCompletionsSUT,
 )
+from dataclasses import dataclass
+from enum import Enum
 from pydantic import BaseModel
 from string import Template
 from typing import Callable, Dict, List, Optional
@@ -17,26 +19,42 @@ class LlamaGuardAnnotation(BaseModel):
     violation_categories: List[str]
 
 
+class LlamaGuardVersion(Enum):
+    VERSION_1 = "Llama Guard 1"
+    VERSION_2 = "Llama Guard 2"
+
+
 class LlamaGuardAnnotator(CompletionAnnotator[LlamaGuardAnnotation]):
     """Send the SUT's completions to LlamaGuard to get safety ratings."""
 
     def __init__(
         self,
         together_api_key: TogetherApiKey,
-        formatter: Optional[Callable[[str, str], str]] = None,
+        formatter: Optional[Callable[[str], str]] = None,
         decoder: Optional[Dict[str, str]] = None,
+        version: LlamaGuardVersion = LlamaGuardVersion.VERSION_2,
     ):
-        self.model = TogetherCompletionsSUT("annotator", _MODEL_NAME, together_api_key)
-        self.formatter = _default_formatter if formatter is None else formatter
-        self.decoder = _DEFAULT_MAPPING if decoder is None else decoder
+        if version == LlamaGuardVersion.VERSION_1:
+            self.model_name = "Meta-Llama/Llama-Guard-7b"
+            formatter = _formatter_v1 if formatter is None else formatter
+        elif version == LlamaGuardVersion.VERSION_2:
+            self.model_name = "meta-llama/LlamaGuard-2-8b"
+            formatter = _formatter_v2 if formatter is None else formatter
+        else:
+            raise ValueError(f"Unknown LlamaGuard version: {version}")
+        self.model = TogetherCompletionsSUT(
+            "annotator", self.model_name, together_api_key
+        )
+        self.formatter = formatter
+        self.decoder = _make_llamaguard_mapping(version) if decoder is None else decoder
 
     def translate_request(
         self, prompt: PromptWithContext, completion: SUTCompletion
     ) -> TogetherCompletionsRequest:
         # TODO Consider giving more than just the agent's response
         return TogetherCompletionsRequest(
-            prompt=self.formatter(_AGENT, f"{_AGENT}: {completion.text}"),
-            model=_MODEL_NAME,
+            prompt=self.formatter(completion.text),
+            model=self.model_name,
             # This might need to be dynamic if the decoder is complicated.
             max_tokens=20,
             n=1,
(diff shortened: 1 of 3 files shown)


wpietri:
I'm excited to take a look. And @bollacker, I think this has several things worth thinking about in terms of versioning.
* The change to LlamaGuard 2 obviously means a version bump, but to what? Modelguard definitely, but does the Test have an independent version? (I think the test-makers think so).
* The change to the version 1 prompt formatting is interesting:
** Does that also bump the test version?
**  To what extent do we need to be able to run with old versions of a Test? Is that always the same as a version of the code, or are there times where we want some code improvements but not the ones that affect a test?
** For example, if we add a new core SUT or improve parallelization, do we ever want to run that with, say, a published version of a benchmark? If so, we may need a specific Test to accept a version parameter and maintain bug compatibility, such that the prompt format changes would be switchable.
** If we do that, how long do we need to carry the ability to run old Tests? I've seen projects maintain backwards compatibility like that and it gets increasingly hard.

None of this should hold up the patch, but I wanted to raise it while we have a good example.

plugins/together/modelgauge/annotators/llama_guard_annotator.py
```diff
@@ -17,26 +19,42 @@ class LlamaGuardAnnotation(BaseModel):
...
-        formatter: Optional[Callable[[str, str], str]] = None,
+        formatter: Optional[Callable[[str], str]] = None,
         decoder: Optional[Dict[str, str]] = None,
+        version: LlamaGuardVersion = LlamaGuardVersion.VERSION_2,
     ):
-        self.model = TogetherCompletionsSUT("annotator", _MODEL_NAME, together_api_key)
-        self.formatter = _default_formatter if formatter is None else formatter
-        self.decoder = _DEFAULT_MAPPING if decoder is None else decoder
+        if version == LlamaGuardVersion.VERSION_1:
+            self.model_name = "Meta-Llama/Llama-Guard-7b"
```
wpietri, line 38: This might be worth pushing into the enum.

plugins/together/modelgauge/annotators/llama_guard_annotator.py
```diff
@@ -66,86 +84,220 @@ def translate_response(
...
+        LLAMA_GUARD_2_CATEGORY_PREFIX,
+    )
 
 
-def _default_formatter(role, message):
-    prompt = _DEFAULT_TEMPLATE.substitute(prompt=message, agent_type=role)
-    prompt = f"<s>[INST] {prompt.strip()} [/INST]"
-    return prompt
+def _make_llamaguard_mapping(llama_guard_version):
+    if llama_guard_version == LlamaGuardVersion.VERSION_1:
```
wpietri, line 294: This might be another good thing to add to the enum. Generally in OO design, any place you see an 'if Thing1 else Thing2' construct, it's worth looking to see if that can be replaced with polymorphism. Then you can just call `thing.categories`, which simplifies the code.

plugins/together/modelgauge/annotators/llama_guard_annotator.py
```diff
@@ -66,86 +84,220 @@ def translate_response(
...
+        categories = LLAMA_GUARD_1_CATEGORY
+        prefix = LLAMA_GUARD_1_CATEGORY_PREFIX
+    else:
+        categories = LLAMA_GUARD_2_CATEGORY
+        prefix = LLAMA_GUARD_2_CATEGORY_PREFIX
+    return {f"{prefix}{i+1}": c.name for i, c in enumerate(categories)}
```
wpietri, line 300: Is there a test for this? It seems like the string is pretty complicated. If I broke it while working on it, I'd have a hard time spotting it. If you're inspecting it manually, perhaps you could put some of that knowledge into a test or two?

plugins/together/tests/test_llama_guard_annotator.py
```diff
@@ -46,44 +48,73 @@ def evaluate(self, request: TogetherChatRequest) -> TogetherChatResponse:
         return self.response_stack.pop()
 
 
-def test_safe():
+@pytest.mark.parametrize(
```
wpietri, line 51: It's always nice to see test improvements along with code improvements!

wpietri (approved):
Fine as is, but I added some options for improvement if you wanted.

bollacker (approved):
LGTM

bollacker:
> I'm excited to take a look. And @bollacker, I think this has several things worth thinking about in terms of versioning.
> 
> * The change to LlamaGuard 2 obviously means a version bump, but to what? Modelguard definitely, but does the Test have an independent version? (I think the test-makers think so).

Yes.  A change in evaluator is a change to the test.

> * The change to the version 1 prompt formatting is interesting:
>   
>   * Does that also bump the test version?

Yes.  I think so.

>   * To what extent do we need to be able to run with old versions of a Test? Is that always the same as a version of the code, or are there times where we want some code improvements but not the ones that affect a test?



>   * For example, if we add a new core SUT or improve parallelization, do we ever want to run that with, say, a published version of a benchmark? If so, we may need a specific Test to accept a version parameter and maintain bug compatibility, such that the prompt format changes would be switchable.

Yeah.  I can imagine needing to run with an old test.  The question is one of how to refer to the old test.  Would that mean checking out a specific tag in the code?

>   * If we do that, how long do we need to carry the ability to run old Tests? I've seen projects maintain backwards compatibility like that and it gets increasingly hard.

Let's chat about this.  I'm pretty certain folks we'll need to provide some form of access to old tests, but how much and for how long is up for debate.   There are wacky things (e.g. putting the version number as part of the name of the test).  I think using git versioning as a poor man's tool to manage this is probably not sustainable.
 
> None of this should hold up the patch, but I wanted to raise it while we have a good example.

Yeah.  We need use cases to drive.

bkorycki:
You bring up some important questions about versioning that definitely need further discussion soon.
> The change to LlamaGuard 2 obviously means a version bump, but to what? Modelguard definitely, but does the Test have an independent version? (I think the test-makers think so).

Yeah, I see value in versioning tests independently. 
Linking test versions directly to ModelGauge could lead to some confusion. For example, every version bump in ModelGauge might suggest changes to the tests, potentially undermining trust in the previous test results. Which becomes more problematic with frequent patching that doesn’t actually affect a test, yet the version suggests otherwise.
> Does that also bump the test version?

I think it should. I see 3 areas of consideration when it comes to bumping a test version
1. Changes the underlying test data, which is already sort of handled by DependencyHandler
2. Changes to it’s own logic/functionality that might change results 
3. Changes to plugins that it depends on (e.g. LlamaGuard).

plugins/together/modelgauge/annotators/llama_guard_annotator.py
bkorycki: Updated with your feedback, thanks!

plugins/together/modelgauge/annotators/llama_guard_annotator.py
bkorycki: I think the only way you could break it is if the decoder's categories are not aligned with the categories used in the prompt formatter. And any misalignment (e.g. prompt formatter uses category `A1: Hate` and decoder uses `B1: Hate` would be caught by the existing tests. Are there any other potential ways to mess stuff up here that you foresee here?

PR #434, opened 2024-05-01
Together ProducesPerTokenLogProbabilities capability

bkorycki:
This PR adds the ProducesPerTokenLogProbabilities capability to all three Together SUTs (completion, chat, inference).
It also adds more comprehensive testing for the Together client.

### DIFF
--- a/plugins/together/modelgauge/suts/together_client.py
+++ b/plugins/together/modelgauge/suts/together_client.py
@@ -2,8 +2,18 @@ import requests
 from modelgauge.prompt import ChatPrompt, ChatRole, SUTOptions, TextPrompt
 from modelgauge.prompt_formatting import format_chat
 from modelgauge.secret_values import InjectSecret, RequiredSecret, SecretDescription
-from modelgauge.sut import PromptResponseSUT, SUTCompletion, SUTResponse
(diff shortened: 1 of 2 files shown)


yifanmai (approved):
Awesome, thanks!

wpietri (approved):
This looks great. One case I wasn't quite sure about: does Together return logprobs for every model? If so, that's great. If not, what happens if somebody asks for logprobs but together doesn't return them?

bkorycki:
> does Together return logprobs for every model? If so, that's great.

I think so. It did for all the models I tested. And Together doesn't have any disclaimers about logprobs being limited to certain models.

> If not, what happens if somebody asks for logprobs but together doesn't return them?

Then it would fail on the first call to `evaluate()`. Which is not ideal, but it does effectively signal to a user at the beginning of a test run that the model cannot be used for tests that require logprobs.

yifanmai:
The way Together seems to work is that they run a different HTTP server in front of each GPU instance, and the HTTP servers can be of different versions and have slightly different APIs... I have encountered missing logprobs in some server versions, but I think the assertion is sufficient to catch any future regressions.

PR #439, opened 2024-05-16
Add LogProbs capability to HuggingFaceSUT

bkorycki:
This cleans up some code left over from old helm and formally adds the `ProducesPerTokenLogProbabilities` capability to HuggingFaceSUT. I've also made a few subtle changes to how generation parameters are handled, aiming to avoid unintentional side effects.
- Don't pass a `generation_config` parameter. 
Instead, specify `pad_token_id` directly as a kwarg. Passing a generation config can have unintended consequences by overriding the model's default generation config, including parameters that we were not trying to change.
- Consistent handling of stop sequences.
Stop_sequences are always passed via the `stopping_criteria` parameter. Previously, single-token stop sequences were passed by setting `eos_token_id` instead, which leads to inconsistencies by altering the model's default configuration without clear disclosure to users.
- Pass `top_k` (if it was specified by the prompt).

### DIFF
--- a/plugins/huggingface/modelgauge/suts/huggingface_client.py
+++ b/plugins/huggingface/modelgauge/suts/huggingface_client.py
@@ -148,8 +148,6 @@ class HuggingFaceToken(OptionalSecret):
     ]
 )
 class HuggingFaceSUT(PromptResponseSUT[HuggingFaceRequest, HuggingFaceResponse]):
-    """A thin wrapper around a Hugging Face AutoModelForCausalLM for HuggingFaceClient to call."""
-
     def __init__(
         self,
         uid: str,
--- a/plugins/huggingface/modelgauge/suts/huggingface_client.py
+++ b/plugins/huggingface/modelgauge/suts/huggingface_client.py
@@ -11,7 +11,7 @@ from modelgauge.sut import PromptResponseSUT, SUTCompletion, SUTResponse
 from modelgauge.sut_capabilities import AcceptsChatPrompt, AcceptsTextPrompt
 from modelgauge.sut_decorator import modelgauge_sut
 from modelgauge.sut_registry import SUTS
-from pydantic import BaseModel
+from pydantic import BaseModel, ConfigDict, field_serializer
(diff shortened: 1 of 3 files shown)


plugins/huggingface/modelgauge/suts/huggingface_client.py
```diff
@@ -290,130 +199,78 @@ def evaluate(self, raw_request: HuggingFaceRequest) -> HuggingFaceResponse:
...
-            all_generated_tokens_logprobs,
-            all_generated_tokens_top_logprobs_dicts,
-        ):
+        for completion_id in range(raw_request.generate_args.num_return_sequences):
+            # Remove input prompt and truncate if length exceeds max_new_tokens.
+            sequence = output.sequences[completion_id][num_input_tokens:]
+            sequence = sequence[:max_new_tokens]
+
+            text = tokenizer.decode(sequence)
+            sequence_logprobs_dicts = None
```
dhosterman, line 249: Is it possible to just initialize this with an empty list instead of having it be `None` here and an empty (or not) list later?

bkorycki: So the idea behind the ProducesPerTokenLogProbs capability is that logprobs should only be returned if they are specifically requested. And I think `None` makes it explicit that the completion does not include logprobs, whereas an empty list is more ambiguous (e.g. Were logprobs not requested or was there an error in getting logprobs that caused to be empty?)

dhosterman (approved):
👍

wpietri (approved):
This looks great and I love the thorough testing. But I must confess I don't understand all the issues driving the choices. Perhaps once things settle down you could give me and Daniel a bit of a tour of this stuff?

bkorycki:
Yes that's a good idea! I can do that

yifanmai (approved):
This is great; thanks for the extensive testing!

I will confess that I do not know all the ins-and-outs of the Hugging Face `generate()` API, which is really complicated -_- but I have medium to high confidence that what you're doing is correct.

PR #448, opened 2024-06-10
Add remote attach debug option compatible with `poetry run`

tsunamit:
Allows for debugger attaching with debugpy (If using VSCode, will need to add an attach config in launch.json)
- Add `--debug` flag when running a test with poetry
- Test will pause before starting, allowing debugger to attach
- If you're using VSCode, you'll need to add a remote attach config to your launch.json. Use this guide: https://code.visualstudio.com/docs/editor/debugging#_launch-versus-attach-configurations
- The guide I used to find this solution: https://github.com/python-poetry/poetry/issues/5354#issuecomment-1852884176

### DIFF
--- a/modelgauge/main.py
+++ b/modelgauge/main.py
@@ -1,5 +1,6 @@
 import os
 import click
+import debugpy
 from modelgauge.base_test import PromptResponseTest
 from modelgauge.command_line import (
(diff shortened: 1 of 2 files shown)


wpietri:
Is this still an active pull request?

PR #450, opened 2024-06-18
Add batch test runner under preview flag

tsunamit:


bkorycki:
Yeah I think this approach is sound! 
I wonder if it would make sense to build dynamically-sized batches of test items (such that their total prompt count < `max_batch_size`) instead of prompt batches. And implement batch versions of translate_prompt/response(), either in the PromptResponseSUT class or just in the runner. This would simplify the `translate_prompt`, `evaluate`, `translate_response` execution sequence quite a bit. I don’t know how much that's worth the trouble though. But good work so far!

tsunamit:
To be honest, this is probably less complicated if I can expose an async interface instead of batch. Then we can keep the logic much simpler (single task). Furthermore, most of our endpoints likely use async rather than batch.

PR #454, opened 2024-06-25
Document interface for SUTs

bkorycki:
API reference for SUTs using mkdocs (already used for readthedocs) and the mkdocstrings plugins.
The documentation is loaded from docstrings and type hints.

https://github.com/mlcommons/modelgauge/assets/12498420/4316fc48-f754-4cbf-b4ab-e58ddae4ce0f


dhosterman (approved):
Looks great!

wpietri (approved):

PR #465, opened 2024-07-09
Add multithreaded `run-prompts` command and supporting code

wpietri:
This adds a `modelgauge run-prompts` command to aid workstreams in quickly running batches of prompts not currently associated with tests. It will handle multiple SUTs in the same run. It is multithreaded with controllable concurrency, so that users can experiment to find the fastest way to run their particular workload.

It also adds Pipeline and related classes, which I hope will serve as the base for a composable set of objects that handle common bulk processing tasks like running prompts, getting annotations, and any other slow I/O-bound workloads.

 I'd especially love feedback on the Pipeline stuff, especially use conventions, method and class names, and the docs.  It all works, but I have stared into the concurrency abyss too long to be sure that it makes sense to its intended users.


modelgauge/main.py
```diff
@@ -228,6 +241,81 @@ def run_test(
...
+    "-s",
+    "--sut",
+    help="Which registered SUT(s) to run.",
+    multiple=True,
+    required=True,
+)
+@click.option(
+    "--workers",
+    type=int,
+    default=None,
```
tybrs, line 256: This gives an error if no `--workers` arg is given because `thread_count` is inherited from workers. Maybe we should probably give this a different default value or make it a required argument.

```shell
poetry run modelgauge run-prompts id_draft_id_elec_2024-06-28.csv --sut gemma-7b-it --sut Mixtral-8x7B-Instruct-v0.1 --sut llama-3-70b-chat-hf
```
```python
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/localdisk/twilbers/src/repos/modelgauge/modelgauge/main.py", line 324, in main
    modelgauge_cli()
  File "/localdisk/twilbers/src/repos/modelgauge/.venv/lib/python3.10/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
  File "/localdisk/twilbers/src/repos/modelgauge/.venv/lib/python3.10/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
  File "/localdisk/twilbers/src/repos/modelgauge/.venv/lib/python3.10/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/localdisk/twilbers/src/repos/modelgauge/.venv/lib/python3.10/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/localdisk/twilbers/src/repos/modelgauge/.venv/lib/python3.10/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
  File "/localdisk/twilbers/src/repos/modelgauge/modelgauge/main.py", line 314, in run_prompts
    pipeline.run()
  File "/localdisk/twilbers/src/repos/modelgauge/modelgauge/pipeline.py", line 288, in run
    segment.start()
  File "/localdisk/twilbers/src/repos/modelgauge/modelgauge/pipeline.py", line 168, in start
    for i in range(self.thread_count):
```

tsunamit:
The pipeline approach + python threading is feasible here and I see how it works.

Was curious though, was there a reason for not using asyncio? I believe asyncio is typically used for managing slow I/O bound jobs + several connections. My guess is perhaps you don't need the granularity of asyncio (indicating exactly where to await).

I gave this a few tries as well - had a couple (poor) implementations to help manage my batch LLM workloads. This does seem cleaner, thanks for doing it.

modelgauge/main.py
wpietri: Ah, great catch. This used to work; I'll make it work again.

wpietri:
> The pipeline approach + python threading is feasible here and I see how it works.
> 
> Was curious though, was there a reason for not using asyncio? I believe asyncio is typically used for managing slow I/O bound jobs + several connections. My guess is perhaps you don't need the granularity of asyncio (indicating exactly where to await).

Thanks for the feedback Ryan!

I went this route mainly because I'm more familiar with threaded approaches, but it also lets us take better advantage of multi-CPU machines. I was also more confident that I could hide most of the threading machinery and present something simple to use. Here people can just write a few subclasses with methods that don't have to know how to cooperate, but at least given my lower async/await experience, I didn't know how to keep devs from having to write proper async/await stuff to handle items and do work. 


> I gave this a few tries as well - had a couple (poor) implementations to help manage my batch LLM workloads. This does seem cleaner, thanks for doing it.

Glad to hear it! Since it sounds like you might use it soon, any feedback on class names, method names, or use conventions? If there's something that seems weird or puzzling, I'd be happy to see what I can do.

tsunamit:
Wanted to preface first that I agree with the current approach and look forward to trying to use it for the annotator runner.

However, a small detail regarding the following - 
> I went this route mainly because I'm more familiar with threaded approaches, but it also lets us take better advantage of multi-CPU machines.

 iirc the python global interpreter lock (GIL) limits even the threading library to 1 executing thread at once (so we don't take advantage of multiple CPUs). If we're worried about being CPU bound, then perhaps the multiprocessing library is a better fit as that actually enables multi-cpu workers.

From the python docs:
> CPython implementation detail: In CPython, due to the [Global Interpreter Lock](https://docs.python.org/3/glossary.html#term-global-interpreter-lock), only one thread can execute Python code at once (even though certain performance-oriented libraries might overcome this limitation). If you want your application to make better use of the computational resources of multi-core machines, you are advised to use [multiprocessing](https://docs.python.org/3/library/multiprocessing.html#module-multiprocessing) or [concurrent.futures.ProcessPoolExecutor](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ProcessPoolExecutor). However, threading is still an appropriate model if you want to run multiple I/O-bound tasks simultaneously.

That all being said, I think we're mostly worried about being network I/O bound. So multithreading seems to be the right fit. Just wanted to highlight that this implementation will still only use 1 CPU (we just swap threads when blocked on I/O, granting us our desired speedup).

> Glad to hear it! Since it sounds like you might use it soon, any feedback on class names, method names, or use conventions? If there's something that seems weird or puzzling, I'd be happy to see what I can do.

Will do, I think a good test will be to try the evaluator/annotator runners using this. Once I get my other PR checked in I will try with this.

tsunamit:
> > > I went this route mainly because I'm more familiar with threaded approaches, but it also lets us take better advantage of multi-CPU machines.
> > 
> > 
> > iirc the python global interpreter lock (GIL) limits even the threading library to 1 executing thread at once (so we don't take advantage of multiple CPUs).
> 
> That's not quite correct. The GIL limits the interpreter running Python bytecode in different threads, but more goes on. It doesn't constrain anything lower down, like the work that happens in C code and in system calls. If you run this with a sizable job and watch CPU usage, you'll see that the process is indeed using more than 1 CPU's worth of power. Some of which will be threading overhead, but some of it will be productive work. And yes, we'll definitely look at using the multiprocessing stuff it that's insufficient.

Yes limiting python bytecode in different threads is an important distinction - thanks for clarifying.

dhosterman (approved):
This looks solid! Nice work!

tsunamit (approved):
Thanks for doing this - will try to use this with the annotator runner

modelgauge/pipeline.py
```diff
@@ -0,0 +1,310 @@
...
+
+
+class Pipeline:
+    def __init__(
+        self,
+        *segments: PipelineSegment,
+        debug: bool = False,
+        progress_callback: Optional[Callable] = None,
+    ):
+        super().__init__()
```
bkorycki, line 262: I may be wrong but I don't think this is necessary.

modelgauge/main.py
```diff
@@ -228,6 +241,81 @@ def run_test(
     print("Full TestRecord json written to", output_file)
 
 
+@modelgauge_cli.command()
+@click.option(
+    "sut_names",
```
bkorycki, line 246: This is nit-picky buy I think `sut_uids` would be more accurate

modelgauge/main.py
```diff
@@ -228,6 +241,81 @@ def run_test(
...
+    secrets = load_secrets_from_config()
+
+    try:
+        suts = {
+            sut_name: SUTS.make_instance(sut_name, secrets=secrets)
+            for sut_name in sut_names
+        }
+    except MissingSecretValues as e:
+        raise_if_missing_from_config([e])
+
```
bkorycki, line 279: It might be helpful to check that the SUTs all have the `AcceptsTextPrompt` capability before starting the pipeline.

modelgauge/pipeline.py
```diff
@@ -0,0 +1,310 @@
...
+        for s in self._segments:
+            s._debug_enabled = debug
+
+        assert isinstance(self.source(), Source)
+        assert isinstance(self.sink(), Sink)
+
+        for a, b in zip(segments[:-1], segments[1:]):
+            b.set_upstream(a)
+
+    def source(self):
```
bkorycki, line 276: I'm curious why you chose to make this and `sink` methods instead of attributes or `@property` methods.

modelgauge/pipeline.py
```diff
@@ -0,0 +1,310 @@
...
+            timeout = PipelineSegment.default_timeout
+        if self._upstream is None:
+            raise (ValueError(f"{self} doesn't have an upstream connection"))
+        self._debug("getting from upstream")
+
+        try:
+            item = self._upstream._queue.get(timeout=timeout)
+        except queue.Empty:
+            self._debug("get was empty")
+            raise
```
bkorycki, line 95: I'm curious why would this be an error

bkorycki (commented):
You did a great job on this! The design feels intuitive, and I appreciate how well-documented it is. 
My primary suggestion is to use the existing `PromptWithContext` (with `source_id=row["UID"]`) instead `PromptItem`. That way we could re-use some of the pipes in other runners.

modelgauge/prompt_pipeline.py
```diff
@@ -0,0 +1,152 @@
...
+
+    def handle_item(self, item: PromptItem):
+        for sut_key in self.suts:
+            self.downstream_put((item, sut_key))
+
+
+class PromptSutWorkers(Pipe):
+    def __init__(self, suts: dict[str, PromptResponseSUT], workers=None):
+        if workers is None:
+            workers = 8
```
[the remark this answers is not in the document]
wpietri: Because Tyler had an issue if he didn't specify a --workers argument at the command line. That CLI option defaults to `None` currently, and so a `None` was getting passed in here, which was then blowing up, because passing in a `None` overrides whatever default param value was set.

I could have fixed it with more code in the CLI, but since I wanted these objects to be pretty friendly to novices, I figured I'd make this harder to use wrongly.

modelgauge/pipeline.py
wpietri: This is one of those cases where people are using exceptions for things that are not errors. If you call a Queue with a timeout and nothing arrives, it will raise `queue.Empty` as part of its normal operations. Personally, I would have it just return `None`, but I think they did it this way because they wanted people to be able to put `None` into the queue, or at least not to have to worry if they were putting `None` in. 

Here I want to have a good debug message in this case, so I did the catch-and-raise thing. One layer up there's something that catches `queue.Empty` and handles it properly.

modelgauge/pipeline.py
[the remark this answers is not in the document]
wpietri: It's true that Object's init doesn't do anything, so it's true that it's technically optional. But I think it's generally a good habit to have, and since so many of the nearby` __init__` methods have it, I left it in here.

modelgauge/pipeline.py
wpietri: I think it would be weird to have them as direct attributes, as it would be duplicative of the array in a way that could cause problems with future changes. I'd be find with making them properties; though.

Generally I think Python's use of attributes, methods, and properties is a little confused. It exposes implementation details in ways I don't much like, and I think other languages do it better. And the confusion means that different code bases will do different things in cases like this. Maybe we should all chat about it and decide when we want to use properties?

modelgauge/main.py
wpietri: Excellent nit-pick! Fixed.

tybrs (approved):

modelgauge/main.py
wpietri: This was surprisingly hard to test, but I believe I now have this.

bkorycki: Looks good, thanks!

modelgauge/main.py
bkorycki: Thanks!

modelgauge/pipeline.py
bkorycki: Okay I think I understand now, thanks for explaining!

bkorycki (approved):

PR #469, opened 2024-07-11
safe-tests-use-modellab

dhosterman:
Updating to use our current dev deployment of modellab to serve the files for safe tests. I'm not sure this is what we want to do! It might introduce some volatility to the tests and we might wind up reconsidering how we want to store files, but this serves as a reasonable proof of concept that it can work the way we expect it to.


bkorycki:
I'm personally fine with switching over to modellab for file hosting! If we do choose to go ahead with this, I think it would be a good idea to explicitly set the default dependency version requirement for SafeTest to some stable one. Right now the runner always uses the latest version, which is probably not desirable with test volatility. Although we can figure that out later.

bkorycki (approved):

wpietri (approved):
🤠

PR #472, opened 2024-07-12
Adds an optional cache to run-prompts

wpietri:
You can now do things like this:

```
time modelgauge run-prompts  --sut gemma-7b-it --sut Mixtral-8x7B-Instruct-v0.1 --sut llama-3-70b-chat-hf --cache mycache  --workers 24 id_draft_id_elec_2024-06-28.csv
```

That will create the directory mycache and save all the responses there. The second time you run it, it's way faster.

I made it optional with an explicit cache dir so that cache management was firmly in the hands of the users. With more use, we could look at making it built in.

Note: for some reason this doesn't cache like 0.1% of the items on the first run. No idea why, but I'll look into it on Monday. But it's usable as is.

### DIFF
--- a/modelgauge/main.py
+++ b/modelgauge/main.py
@@ -1,9 +1,10 @@
 import datetime
 import os
 import pathlib
-from typing import List, Optional
+from typing import List, Optional, Any, Dict
 
 import click
+from click._termui_impl import ProgressBar
 
 from modelgauge.base_test import PromptResponseTest
 from modelgauge.command_line import (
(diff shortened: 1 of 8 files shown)


bkorycki:
Why not use our caching mechanism? `SqlDictCache`

wpietri:
> Why not use our caching mechanism? `SqlDictCache`

It's slower (which probably doesn't matter much) and harder to use (which does).

bkorycki (approved):

wpietri:
As mentioned in #engineering-support, it turns out the caching was fine. But entries with duplicate texts could both be in fight at once, getting different answers. But on a second cached run, both would get the same answer from the cache. I've redone it so that the cache key is prompt UID plus text.

I also extracted the caching behavior into a superclass, CachingPipe, so that anybody wanting similar caching with work being done in parallel should have an easier time of it.

dhosterman (approved):
Mystery solved! Nice work. This looks great and easy to use.

PR #475, opened 2024-07-17
Annotator pipeline

bkorycki:
This largely follows the structure of the prompt pipeline.


demo_plugin/modelgauge/tests/demo_04_using_annotation_test.py
```diff
@@ -56,7 +56,7 @@ def make_test_items(self, dependency_helper: DependencyHelper) -> List[TestItem]
 
     def get_annotators(self) -> Mapping[str, Annotator]:
         # Request the `DemoYBadAnnotator` be run, and key its results with "badness" for later.
-        return {"badness": DemoYBadAnnotator()}
+        return {"badness": DemoYBadAnnotator("badness_annotator")}
```
tsunamit, line 59: default uids would be nice. does uid still need to be the first parameter?

plugins/openai/modelgauge/annotators/openai_compliance_annotator.py
```diff
@@ -19,12 +19,14 @@ class OpenAIComplianceAnnotator(CompletionAnnotator[ComplianceAnnotation]):
 
     def __init__(
         self,
+        uid="openai_compliance_annotator",
         formatter: Optional[Callable[[PromptWithContext, SUTCompletion], str]] = None,
```
tsunamit, line 23: ah yes, default uids like this seems better

plugins/together/modelgauge/annotators/llama_guard_annotator.py
```diff
@@ -225,11 +227,13 @@ class LlamaGuardAnnotator(CompletionAnnotator[LlamaGuardAnnotation]):
 
     def __init__(
         self,
+        uid: str,
         together_api_key: TogetherApiKey,
```
tsunamit, line 231: had the same issue with lg - because the others don't have default values, we cannot supply a default here and we're stuck needing to init every instance with a string uid

modelgauge/main.py
```diff
@@ -321,6 +330,82 @@ def show_progress(data):
...
+        }
+    except MissingSecretValues as e:
+        raise_if_missing_from_config([e])
+
+    path = pathlib.Path(filename)
+    input = CsvAnnotatorInput(path)
+
+    output_path = path.parent / pathlib.Path(
+        path.stem
+        + "-responses-"
```
tsunamit, line 375: perhaps "annotations"?

modelgauge/annotation_pipeline.py
```diff
@@ -0,0 +1,163 @@
...
+class CsvAnnotatorInput(AnnotatorInput):
+    def __init__(self, path):
+        super().__init__()
+        self.path = path
+
+    def __iter__(self) -> Iterable[AnnotatorInputSample]:
+        with open(self.path, newline="") as f:
+            csvreader = csv.DictReader(f)
+            for row in csvreader:
+                prompt = PromptWithContext(
```
tsunamit, line 50: prompt, response as separate columns fine for now. multiturn out of scope for now, but this may need to change in the futreu

tsunamit (approved):
Looks good to me. Once you get this in I'll add the new annotators from the old PR to the main branch.

demo_plugin/modelgauge/tests/demo_04_using_annotation_test.py
bkorycki: It does unfortunately :( I agree though, default uids would be nice, we can think about it some more later on.

modelgauge/main.py
bkorycki: Good catch, updated!

wpietri (approved):
Looks solid to me. I didn't know about the jsonlines library, which is cool. And nice to see solid tests.

I didn't actually run it, but if you and Ryan are happy with the output, it's good by me.

PR #480, opened 2024-07-19
create safety_models plugin folder and add llama_3_70b_chat teacher model annotator

tsunamit:
- create safety_models plugin folder
- adds llama_3_70b_chat as an annotator


bkorycki:
Would it be possible for working group 3 to just use the `--plugin-dir` command line option when running modelgauge on annotators being worked on? I'm not sure that this needs to be a part of modelgauge.

PR #481, opened 2024-07-23
SUT response + annotation pipeline

bkorycki:
Merges `run-prompts` and `run-annotators` commands into `run-csv-items`.

- `run-csv-items -s sut_1 -s sut_2 prompts.csv`: Get the suts' responses to the prompts the in the input file. 
    - Input CSV headers: UID, Text
    - Output (CSV) headers: UID, Text, sut_1, sut_2
      - SUT columns contain responses
- `run-csv-items -s sut_1 -s sut_2 -a annotator_1 -a annotator_2 prompts.csv` Get the suts' responses to the prompts and then annotate the prompt + responses.
    - Input CSV headers: UID, Text
    - Output: Same as the output below.
- `run-csv-items -a annotator_1 -a annotator_2 prompt_responses.csv` Get annotations for prompt + responses in the input file.
    - Input CSV headers: UID, Prompt, SUT, Response
    - Output: JSON lines file. Each json object has the form:
        ```
        {
          "UID": ..., "Prompt": ..., "SUT": ..., "Response": ..., 
          "Annotations": {"annotator_1": ..., "annotator_2": ...}
        {
       ```

### DIFF
--- a/modelgauge/main.py
+++ b/modelgauge/main.py
@@ -4,16 +4,7 @@ import os
 import pathlib
 from typing import List, Optional
 
-
 from modelgauge.annotator_registry import ANNOTATORS
-from modelgauge.annotation_pipeline import (
-    AnnotatorAssigner,
-    AnnotatorSink,
-    AnnotatorSource,
-    AnnotatorWorkers,
-    CsvAnnotatorInput,
-    JsonlAnnotatorOutput,
-)
 from modelgauge.base_test import PromptResponseTest
 from modelgauge.command_line import (
     DATA_DIR_OPTION,
@@ -33,16 +24,8 @@ from modelgauge.dependency_injection import list_dependency_usage
 from modelgauge.general import normalize_filename
 from modelgauge.instance_factory import FactoryEntry
 from modelgauge.load_plugins import list_plugins
-from modelgauge.pipeline import Pipeline
+from modelgauge.pipeline_runner import run_annotator_pipeline, run_prompts_pipeline
 from modelgauge.prompt import SUTOptions, TextPrompt
-from modelgauge.prompt_pipeline import (
-    PromptSource,
-    PromptSutAssigner,
-    PromptSutWorkers,
-    PromptSink,
-    CsvPromptInput,
-    CsvPromptOutput,
-)
 from modelgauge.secret_values import MissingSecretValues, RawSecrets, get_all_secrets
 from modelgauge.simple_test_runner import run_prompt_response_test
 from modelgauge.sut import PromptResponseSUT, SUT
@@ -258,7 +241,7 @@ def run_test(
     "--sut",
     help="Which registered SUT(s) to run.",
     multiple=True,
-    required=False,
+    required=True,
 )
 @click.option(
     "annotator_uids",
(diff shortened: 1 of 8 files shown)


modelgauge/main.py
```diff
@@ -280,123 +287,82 @@ def run_test(
 )
 @click.argument(
     "filename",
-    type=click.Path(exists=True),
+    type=click.Path(exists=True, path_type=pathlib.Path),
 )
-def run_prompts(sut_uids, workers, cache_dir: pathlib.Path, debug, filename):
-    """Take a CSV of prompts and run them through SUTs. The CSV file must have UID and Text columns, and may have others."""
-
+def run_prompts(sut_uids, annotator_uids, workers, sut_cache_dir, debug, filename):
```
wpietri, line 292: This function is long and hard to follow. Could it be broken up some, either by extracting functions or by creating some sort of state object or runner object?

modelgauge/prompt_pipeline.py
```diff
@@ -1,14 +1,25 @@
...
 
 from modelgauge.pipeline import Source, Pipe, Sink, CachingPipe
 from modelgauge.prompt import TextPrompt
 from modelgauge.single_turn_prompt_response import PromptWithContext
-from modelgauge.sut import PromptResponseSUT, SUT
+from modelgauge.sut import PromptResponseSUT, SUT, SUTCompletion
+
+
+@dataclass
+class SutInteraction:
```
wpietri, line 16: Nice. Typing helps clarify things, and I like the name especially.

tests/test_annotation_pipeline.py
```diff
@@ -182,3 +197,47 @@ def track_progress(data):
 
     assert progress_items[0]["completed"] == 0
     assert progress_items[-1]["completed"] == 4
+
+
+@pytest.mark.parametrize(
+    "sut_worker_count,annotator_worker_count",
+    [(1, 1), (2, 2), (8, 8), (1, 5), (5, 1), (3, 9), (9, 3)],
```
wpietri, line 204: I like this too.

tests/test_cli.py
```diff
@@ -92,7 +96,7 @@ def test_run_prompts_normal(tmp_path):
     assert result.exit_code == 0
 
     out_path = re.findall(r"\S+\.csv", result.stdout)[0]
-    with open(in_path.parent / out_path, "r") as f:
+    with open(tmp_path / out_path, "r") as f:
```
wpietri, line 99: Ah, nice cleanup. Thanks.

wpietri (changes requested):
Looks solid except for the CLI method being quite large and hard to follow.

modelgauge/main.py
bkorycki: That's fair, it was pretty long. I just updated it. Does it look better now?

modelgauge/prompt_pipeline.py
bkorycki: Yay!

modelgauge/main.py
wpietri: I think this is getting closer, but I think it's still confusing. Some of what I think of as the CLI work, like validating arguments, printing user messages, and progress bars, is now off in another file. The functional decomposition with descriptive names does help convey more of the intent. But the functions have a lot of arguments, many in common. And they have a fair bit of branching, including spots with similar code, like in the `if run_annotators` block.

Instead of a functional breakdown, maybe a PipelineRunner object would help? Happy to sketch something if you'd like.

wpietri (changes requested):
I guess for GH ritual reasons I should say: good progress, needs to go further, please see the long comment that I perhaps should have put here.

modelgauge/main.py
bkorycki: Do you imagine `PipelineRunner` to handle all 3 types of pipelines?

wpietri: I'm thinking so. When you have a bunch of things that keep appearing in function signatures together, that's a sign that there might be an object waiting to be named, where those are the fields of the object. When there is code that goes with them, that could end up as methods on the object. Seems like there's some common stuff to all three pipelines!

bkorycki: Okay, I put the logic into a runner object. What do you think?

wpietri (approved):
Great work! This seems much improved to me. I can now easily tell which bits go with which case, and I like the separation of concerns between the runner object and the CLI itself. How does it feel to you?

PR #483, opened 2024-07-25
Proxy server

wpietri:
A simple proxy server for interviewing.


modelgauge/api_server.py
```diff
@@ -0,0 +1,104 @@
...
+from pydantic import BaseModel
+
+from modelgauge.annotator import CompletionAnnotator
+from modelgauge.annotator_registry import ANNOTATORS
+from modelgauge.config import load_secrets_from_config
+from modelgauge.load_plugins import load_plugins
+from modelgauge.prompt import TextPrompt
+from modelgauge.sut import PromptResponseSUT
+from modelgauge.sut_registry import SUTS
+from modelgauge.suts.together_client import CHAT_MODELS
```
bkorycki, line 17: Do you envision this server only being used to interact with Together models? If so, we should probably move this file to the together plugin, because I don't think anything in the core of modelgauge should depend on an optional plugin.

modelgauge/api_server.py
```diff
@@ -0,0 +1,104 @@
...
+
+auth_header = APIKeyHeader(name="x-key")
+
+
+@app.post("/")
+async def postroot(req: ProcessingRequest, key: str = Depends(auth_header)):
+    if key != SECRET_KEY:
+        raise HTTPException(401, "not authorized; send x-key header")
+    if req.annotators:
+        work_items = list(itertools.product(req.prompts, req.suts, req.annotators))
```
bkorycki, line 97: If there are multiple annotators, they might annotate different SUT responses for the same <prompt, sut> pair.  Not sure if this is intentional.

modelgauge/api_server.py
wpietri: Interesting point. Let's discuss this in standup tomorrow. I agree this violates the initial intention of the breakdown for sure, but I'm not sure what to do about that.

modelgauge/api_server.py
wpietri: Could you give an example where that's the case? I'm having trouble imagining it.

bkorycki: Lets say we have `ProcessingRequest("hello", ["some_sut"], ["annotator_1", "annotator_2"])`. This will result in two function calls
- `process_work_item("hello", "some_sut", annotator_1)` 
- `process_work_item("hello", "some_sut", annotator_2)`

These will each call `sut.evaluate(...)` independently. And sometimes suts generate different responses for the same prompt.

modelgauge/api_server.py
bkorycki: Could we just import the entire SUT registry, like how the file imports the ANNOTATORS registry?

bkorycki (changes requested):
Looks good!

pyproject.toml
```diff
@@ -97,10 +99,9 @@ build-backend = "poetry.core.masonry.api"
...
 standard_tests = ["modelgauge_standard_tests"]
 openai = ["modelgauge_openai"]
-together = ["modelgauge_together"]
 huggingface = ["modelgauge_huggingface"]
 perspective_api = ["modelgauge_perspective_api"]
-all_plugins = ["modelgauge_demo_plugin", "modelgauge_openai", "modelgauge_huggingface", "modelgauge_standard_tests", "modelgauge_perspective_api", "modelgauge_together"]
+all_plugins = ["modelgauge_demo_plugin", "modelgauge_openai", "modelgauge_huggingface", "modelgauge_standard_tests", "modelgauge_perspective_api"]
 
 
 [[tool.poetry_bumpversion.replacements]]
```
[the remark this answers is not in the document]
wpietri: Great point. Fixed!

modelgauge/annotators/README.md
```diff
@@ -1,5 +1,5 @@
 # Annotator plugins
 
-ModelGauge uses [namespace plugins](../../docs/plugins.md) to separate the core libraries from the implementations of specific Annotators. That way you only have to install the dependencies you actually care about.
+ModelGauge uses [namespace plugins](../../docs/plugins.md) to separate the core libraries from the implementation of less central code. That way you only have to install the dependencies you actually care about.
```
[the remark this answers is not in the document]
wpietri: Also fixed.

modelgauge/api_server.py
wpietri: Ooh, just saw this. That's a great point. Let me see what I can do.

bkorycki (approved):

PR #485, opened 2024-07-29
Annotator pipeline uses cache

bkorycki:


wpietri (approved):
Nice! I especially like the thorough testing. I'm getting a little suspicous of the pipeline stuff, though. If we need to work with it again, let's spend a little time looking for possible improvements.

bkorycki:
Agreed!

PR #512, opened 2024-08-12
Fix tests for package based installation

rogthefrog:
It looks like the package was being installed from pypi, rather than the local code that we are testing.


wpietri (approved):
This all looks great to me. Thanks for running it down.

PR #529, opened 2024-08-27
Remove safetest base class

bkorycki:
This PR re-organizes SafeTest's versions.  This is motivated by my experience last week in attempting to update SafeTest v1.0 with a new annotator for the demo. The rigidity of sharing a `SafeTestBase` parent made it much more difficult to update an individual test version. Plus I realized that the two versions will actually end up having very little in common, so we do not gain much from the base class.
So I would like to de-tangle the two versions from one another in preparation for a more elegant annotator integration this week.

### DIFF
--- a/modelgauge/tests/safe.py
+++ b/modelgauge/tests/safe.py
@@ -12,22 +12,7 @@ Version 0.5 hazards and tests:
     - ssh: Suicide & Self-Harm
     - hat: Hate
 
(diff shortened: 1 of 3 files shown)


wpietri (approved):
Ooh, great idea!

dhosterman (approved):
Good stuff!

PR #530, opened 2024-08-27
Minor improvements for pipeline debugging.

wpietri:


bkorycki (commented):
Good idea! I'm not familiar w/ standard practice but does it make sense for all threads for a certain segment to have the same name? Instead of like threadname-1, threadname-2, etc.

wpietri:
> Good idea! I'm not familiar w/ standard practice but does it make sense for all threads for a certain segment to have the same name? Instead of like threadname-1, threadname-2, etc.

Great point! Fixed.

I also cleaned up some issues that happen when a pipe segment has an error, which should hopefully make debugging easier.

bkorycki:
Great thanks!

bkorycki (approved):

PR #534, opened 2024-08-28
Adding 'system' role to openai_client _ROLE_MAP

shachihk-intel:
The OpenAI client SUT under plugins Role_map currently only has User and Assistant roles defined. Using system prompts for OpenAI GPT models require the 'system role'.


bkorycki (approved):
👍

PR #535, opened 2024-08-29
Keep track of items that can't be processed

bkorycki:
- Execution continues if a SUT or annotator raises an exception while processing a test item.
- A list of the unprocessed test items + their errors is included in the JSON report.


wpietri (approved):
Seems like a good first step to resolving this. Once this is merged, I'll add support in the new runner. Guess we also need to do a release soon. And then we'll need to figure out how to bubble this up in the benchmarks.

rogthefrog (approved):
This is nice!

PR #550, opened 2024-09-05
Officially add new annotators

bkorycki:
This PR updates the actual v1.0 test to use new annotators.
- The test specifies the set of individual annotators instead of interfacing with a single ensemble annotator object. This enables finer-grained  caching at the individual annotator level. The ensemble class also ran into caching issues due to polymorphic serialization.
- The test uses ensemble’s aggregation logic during the measurement phase. This requires a tiny change to the other repo.
- Does NOT require any environment variables to be set. yay

Problem:  A core test is relying on a private repo. If you don’t have that secret repo installed, everything breaks.


wpietri (approved):
Great progress for now, and I'm happy to have it merged. For the future, I think we can extract an object there. Having a parallel series of ifs can be a great time to introduce a superclass and then two subclasses, each taking a side of the if.

PR #559, opened 2024-09-10
mypy is a little overzealous

rogthefrog:
Some formatters (e.g. isort) move mypy's `# type: ignore` comment to the end of the parsed line. mypy only understands it if it's at the end of the line in the file. :(


wpietri (approved):
Great idea!

PR #561, opened 2024-09-11
HuggingFace Inference SUT

bkorycki:


wpietri (approved):
Long term I'd like to see much of the HuggingFace shenanigans pulled out elsewhere, but this is fine for now.

bkorycki:
@wpietri Like what specifically? The code all seems specific to huggingface to me.

rogthefrog (approved):
👍🏻 

A few thoughts about how to address the together/huggingface/other client management in [this comment](https://github.com/mlcommons/modelgauge/pull/560#issuecomment-2342405136).

wpietri:
> @wpietri Like what specifically? The code all seems specific to huggingface to me.

That's exactly my concern. There's a lot about Hugging Face that's not particularly about a SUT. Since we have annotators that also deal with HuggingFace, there's duplication to be extracted. I also think the stuff about waiting to spin up a model is very much not about SUT-ness, and more about operational concerns.

